PR #176, opened 2023-09-28
Webhooks + Cancel

farhanW3:
## Changes

- Added webhooks + cancel updates

### Webhooks

Add env variable `WEBHOOK_URL` and the JSON data will be sent on the url

### Cancel Updates

- new endpoint `/transaction/cancel` POST end-point added. 
  - this will cancel the transaction request both on DB & on chain. Synchronous action. 
  - Only Caveat: If the queueId being cancelled is picked by the worker and locked then the cancel update doesn't work due to lock on the data on `transactions` table


src/db/transactions/updateTx.ts
```diff
@@ -21,7 +22,9 @@ export const updateTx = async ({
   txData,
 }: UpdateTxParams) => {
   const prisma = getPrismaWithPostgresTx(pgtx);
-
+  logger.server.debug(
+    `Updating transaction with queueId ${queueId}, status ${status}}`,
```
adam-maj, line 26: For these, we should follow the format we're using elsewhere with the `[Transaction] [${queuId}] Message`

src/db/transactions/updateTx.ts
```diff
@@ -86,5 +90,17 @@ export const updateTx = async ({
         },
       });
       break;
+    case TransactionStatusEnum.Cancelled:
+      logger.server.debug(`Cancelling transaction with queueId ${queueId}`);
```
adam-maj, line 94: Same here with the logging format

src/db/transactions/updateTx.ts
```diff
@@ -86,5 +90,17 @@ export const updateTx = async ({
...
+      await prisma.transactions.update({
+        where: {
+          id: queueId,
+        },
+        data: {
+          // TODO: Should we be keeping track of erroredAt here?
```
adam-maj, line 100: We can delete this `TODO` here

src/prisma/migrations/20230927004337_transactions_cancelled_at_added/migration.sql
```diff
@@ -0,0 +1,5 @@
+-- AlterTable
```
adam-maj, line 1: Any reason this migration is here with these smart wallet fields at the beginning? Should be included in previous ones right - maybe local DB out of sync?

src/utils/env.ts
```diff
@@ -71,6 +71,16 @@ export const env = createEnv({
     RETRY_TX_CRON_SCHEDULE: z.string().default("*/30 * * * * *"),
     MAX_BLOCKS_ELAPSED_BEFORE_RETRY: z.coerce.number().default(10),
     MAX_WAIT_TIME_BEFORE_RETRY: z.coerce.number().default(600),
+    WEBHOOKS_ENABLED: boolSchema("true"),
```
adam-maj, line 74: What would be the reason to make this disable-able

server/utilities/transaction.ts
```diff
@@ -0,0 +1,75 @@
...
+  switch (txData.status) {
+    case TransactionStatusEnum.Cancelled:
+      error = createCustomError(
+        `Transaction already cancelled with queueId ${queueId}`,
+        StatusCodes.BAD_REQUEST,
+        "TransactionAlreadyCancelled",
+      );
+      break;
+    case TransactionStatusEnum.Queued:
+      console.log("inside Mined/Queued");
```
adam-maj, line 35: Ik this branch is WIP - just another reminder to remove these `console.log` before we ship

server/api/transaction/cancel.ts
```diff
@@ -0,0 +1,70 @@
...
+        ...standardResponseSchema,
+        [StatusCodes.OK]: responseBodySchema,
+      },
+    },
+    handler: async (request, reply) => {
+      const { queueId } = request.body;
+      const walletAddress = request.headers[
+        "x-backend-wallet-address"
+      ] as string;
+      request.log.info(`Canceling transaction with queueId ${queueId}`);
```
adam-maj, line 58: Same comment about the `[Transaction]` log format

server/utilities/transaction.ts
```diff
@@ -0,0 +1,75 @@
...
+      break;
+    case TransactionStatusEnum.Mined:
+      console.log("inside Mined");
+      error = createCustomError(
+        `Transaction already mined with queueId ${queueId}`,
+        StatusCodes.BAD_REQUEST,
+        "TransactionAlreadyMined",
+      );
+      break;
+    case TransactionStatusEnum.Processed:
```
adam-maj, line 52: Won't this method never work if a transaction has already been processed or submitted? Since the worker would still have already picked it up and sent the transaction - but here we're updating the DB state to be out of sync with what actually happened

In the scenario where it's already sent, we may have to sent a replacement transaction with the method @arcoraven mentioned of sending 0 to yourself.

src/db/transactions/updateTx.ts
farhanW3: these are debug logs.,I will remove them ventually. But yes, will follow this convention if I use `.info`

src/prisma/migrations/20230927004337_transactions_cancelled_at_added/migration.sql
farhanW3: I will remove this and create a new migration with only `cancelledAt`.

This alter for added as I synced from main sometime back which has this change & my local didn't.

server/utilities/transaction.ts
farhanW3: yes yes. will remove debug & console.log(). as debug will start printing if someone changes the NODE_ENV on their docker image

server/utilities/transaction.ts
farhanW3: In the current implementation, if the transaction is submitted & then we submit cancel, it will have to go through the worker to get executed (in the current implementation). The expectation was that the transaction is taking time and want to cancel it and redo the transaction again.

But if the expectation of `/cancel` is,  we send the cancel transaction with 0 value on the `/cancel` request and not give it a `queueId` then will have to update the flow.

farhanW3: @adam-maj , synced with @arcoraven going to make the `/cancel` action synchronous and not go through the worker flow as we need it to get sent on-chain immediately

src/utils/env.ts
farhanW3: i already do a URL check so this is redundant, will remove. thanks.

src/utils/env.ts
```diff
@@ -67,10 +67,19 @@ export const env = createEnv({
     RETRY_TX_ENABLED: boolSchema("true"),
     MAX_FEE_PER_GAS_FOR_RETRY: z.string().default("55000000000"),
     MAX_PRIORITY_FEE_PER_GAS_FOR_RETRY: z.string().default("55000000000"),
-    MAX_RETRIES_FOR_TX: z.coerce.number().default(3),
+    MAX_RETRIES_FOR_TX: z.coerce.number().default(15),
```
farhanW3, line 70: 3 was way too short, all my transactions were getting retried on mumbai polygon

src/prisma/migrations/20230927004337_transactions_cancelled_at_added/migration.sql
farhanW3: Now that I think of this, if I remove this then I will not have the type change `INTEGER` to `TEXT`. 

migrations -> `20230922090743_user_ops` -> Adds the column `callGasLimit` as `INTEGER`
migrations -> `20230927004337_transactions_cancelled_at_added` -> updates few columns `INTEGER` to `TEXT` & adds `cancelledAt` field

@adam-maj

src/utils/env.ts
adam-maj: We should switch back to 3

server/api/transaction/cancel.ts
```diff
@@ -0,0 +1,86 @@
...
+      examples: ["success, error"],
+    }),
+    message: Type.String({
+      description: "Response Message",
+      examples: ["Transaction cancelled on-chain successfully"],
+    }),
+    transactionHash: Type.Optional(
+      Type.String({
+        description: "Transaction Queue ID",
+        examples: ["9eb88b00-f04f-409b-9df7-7dcc9003bc35"],
```
adam-maj: update

src/utils/env.ts
farhanW3: My bad, I wanted to update something else. Fixed now

PR #200, opened 2023-10-06
adding auth headers to webhook

IDubuque:
## Changes
- added auth header to queuedTxListener webhook
- added auth header to tx-update-listener webhook

## How this PR will be tested
- tested authorization locally with checkouts migration

## Output
- webhook header contains authorization bearer token with THIRDWEB_API_SECRET_KEY

### DIFF
--- a/src/worker/listeners/queuedTxListener.ts
+++ b/src/worker/listeners/queuedTxListener.ts
@@ -37,6 +37,7 @@ export const queuedTxListener = async (): Promise<void> => {
         const response = await fetch(env.WEBHOOK_URL, {
           method: "POST",
           headers: {
+            "Authorization": `Bearer ${env.THIRDWEB_API_SECRET_KEY}`,
             Accept: "application/json",
             "Content-Type": "application/json",
           },
--- a/src/worker/listeners/queuedTxListener.ts
+++ b/src/worker/listeners/queuedTxListener.ts
@@ -34,13 +34,18 @@ export const queuedTxListener = async (): Promise<void> => {
       if (env.WEBHOOK_URL.length > 0) {
         const parsedPayload = JSON.parse(msg.payload);
         const txData = await getTxById({ queueId: parsedPayload.id });
+
+        let headers: { Accept: string, "Content-Type": string, Authorization?: string } = {
+          Accept: "application/json",
+          "Content-Type": "application/json",
+        };
+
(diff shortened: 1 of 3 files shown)


src/worker/listeners/queuedTxListener.ts
```diff
@@ -37,6 +37,7 @@ export const queuedTxListener = async (): Promise<void> => {
         const response = await fetch(env.WEBHOOK_URL, {
           method: "POST",
           headers: {
+            "Authorization": `Bearer ${env.THIRDWEB_API_SECRET_KEY}`,
```
farhanW3, line 40: adding auth make sense and i guess this add is good for now, but, I kept it without auth for now, as I didn't want to use `THIRDWEB_API_SECRET_KEY` env var becoz, if a user wants to use a different value (which is used by their webhook server where it will validate the token) then it will not work. maybe we should add new env variable for this, `WEBHOOK_AUTH_BEARER_TOKEN` or something.

adam-maj: Yea probably makes sense to separate this out

farhanW3: eventually all this setting/config values will be on the DB and will be able to add/update from TW Dashboard

IDubuque: Thanks for reviewing! Makes sense, I updated to optional auth and using the new env var in the latest commit

farhanW3: @IDubuque can you add env var this into `src/utils/env.ts` & make it optional.

IDubuque: @farhanW3 added to the env schema in the latest commit

farhanW3: Thanks. 

We should do some cleanup here (I wanted to do it post web-hook v0 push, so this PR would be ideal) and extract the `webhook` (fetch call) into it's own function. Otherwise, we will have to update the logic at two places.

Let me know, if can do take this up :)

IDubuque: I'll take up these changes in the PR. I'd like to have auth on main before deploying engine with paper checkouts to avoid freely exposing the webhook endpoints.

farhanW3: I will merge this and then do the updates I mentioned above.

PR #323, opened 2023-12-03
Catch failed transactions in worker

adam-maj:
Currently, if any transaction fails to be sent in `processTx`, the entire `prisma.$transaction` will fail, rolling back all transaction processing.

Transactions can fail in simulation as `provide.populateTransaction` calls `estimateGas` internally. We need to catch this case and enable all other transactions to be processed.

### DIFF
--- a/src/worker/tasks/processTx.ts
+++ b/src/worker/tasks/processTx.ts
@@ -178,27 +178,55 @@ export const processTx = async () => {
           }
 
           // Group all transactions into a single batch rpc request
+          let i = 0;
           const rpcRequests = [];
-          for (const i in txsToSend) {
-            const tx = txsToSend[i];
+          const preparedTxs: typeof txsToSend = [];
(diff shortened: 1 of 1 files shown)


src/worker/tasks/processTx.ts
```diff
@@ -178,27 +178,55 @@ export const processTx = async () => {
...
+            } catch (err: any) {
+              logger.worker.warn(
+                `[Transaction] [${
+                  tx.queueId
+                }] Failed to send transaction with error - ${
+                  err?.message || err
+                }`,
+              );
+
+              await updateTx({
```
jnsdls, line 217: can this fail?

adam-maj: In what scenario? This is just making a raw call to postgres via prisma

adam-maj: We're using it below in this file too

PR #340, opened 2023-12-07
fix: API response body schema has no type `number`

### LINKED ISSUE #339: calling the `decimals` function of a contract returns an error
### Explanation
I am trying to call the `decimals` function of an ERC20 contract but the API returns the below error.

Here is the source code that causes the error:
```js
const response = await thirdwebEngine.contract.read(
  'decimals', // function name
  'mumbai', // chain name
  '<ERC20_CONTRACT_ADDRESS>',
  '' // no args
)
```

Here is the API response of the Engine:
```
ERROR: The value of '#/properties/result' does not match schema definition. 
err: {
  "type": "TypeError",
  "message": "The value of '#/properties/result' does not match schema definition.",
  "stack":
      TypeError: The value of '#/properties/result' does not match schema definition.
          at anonymous0 (eval at build (/app/node_modules/fast-json-stringify/index.js:174:23), <anonymous>:63:18)
          at serialize (/app/node_modules/fastify/lib/reply.js:867:12)
          at preserializeHookEnd (/app/node_modules/fastify/lib/reply.js:504:17)
          at preserializeHook (/app/node_modules/fastify/lib/reply.js:488:5)
          at Reply.send (/app/node_modules/fastify/lib/reply.js:191:7)
          at Object.handler (file:///app/dist/server/routes/contract/read/read.js:29:42)
          at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
  "serialization": {
    "url": "/contract/:chain/:contractAddress/read",
    "method": "GET"
  }
}
```

### Possible cause

I think this happens because the API tries to return a `number` type response but the schema only accepts the return type if it's a string, object, or an array.

https://github.com/thirdweb-dev/engine/blob/v0.0.9/src/server/schemas/sharedApiSchemas.ts#L60

### Environment

**Node.js version:** v20.9.0
**Engine version:** v0.0.9

hsynlms:
APIs may return `number` type response. The schema only accepts if the response is in the type of string, object, or array. This causes internal server errors.

This PR fixes #339 

## Changes

Type `number` has been added to the response body schema list.

## How this PR will be tested

- [ ] Run the Engine.
- [ ] Try to make a GET request to the `/contract/:chain/:contractAddress/read` API to call a contract's `decimals` function without arguments.
Request Params: "{"chain":"mumbai","contractAddress":"<ERC20_CONTRACT_ADDRESS>"}"
Request Querystring: "{"functionName":"decimals"}"
Result: Returns 200 with "{ result: 18 }" in the response body.

## Output

```
{ result: 18 }
```

### DIFF
--- a/src/server/schemas/sharedApiSchemas.ts
+++ b/src/server/schemas/sharedApiSchemas.ts
@@ -57,7 +57,7 @@ export const publishedDeployParamSchema = Type.Object({
  */
 const replyBodySchema = Type.Object({
   result: Type.Optional(
-    Type.Union([Type.String(), Type.Object({}), Type.Array(Type.Any())]),
+    Type.Union([Type.Number(), Type.String(), Type.Object({}), Type.Array(Type.Any())]),
   ),
 });
 


adam-maj:
@hsynlms Great catch - thanks for the PR!

PR #372, opened 2024-01-11
ACCESS_CONTROL_ALLOW_ORIGIN into Config

farhanW3:
## Changes

- Moved the ACCESS_CONTROL_ALLOW_ORIGIN into config.
- Exposed Get/Post/Delete end-points to retrieve, add & remove urls from the CORS config.
- Dynamic CORS Urls loading supported. No need to restart server.

NOTE: Both Auth & CORS hooks should be same. Made them both `preHandler`

image

### DIFF
--- a/src/utils/env.ts
+++ b/src/utils/env.ts
@@ -66,9 +66,6 @@ export const env = createEnv({
       ),
     PORT: z.coerce.number().default(3005),
     HOST: z.string().default("0.0.0.0"),
-    ACCESS_CONTROL_ALLOW_ORIGIN: z
-      .string()
-      .default("https://thirdweb.com,https://thirdweb-preview.com"),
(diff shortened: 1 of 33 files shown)


src/utils/env.ts
```diff
@@ -68,9 +68,6 @@ export const env = createEnv({
       ),
     PORT: z.coerce.number().default(3005),
     HOST: z.string().default("0.0.0.0"),
-    ACCESS_CONTROL_ALLOW_ORIGIN: z
```
adam-maj, line 71: Btw since we removed this here, we should add it to the defaults in the `getConfiguration` functions

farhanW3: Added [here](https://github.com/thirdweb-dev/engine/pull/372/commits/a71db82bd2b9eaf5b991fa4e9afedb337e935e85)

PR #381, opened 2024-01-16
Server Updates: Caching, Logging & CORS

farhanW3:
### Caching
- Added process to clear cache based on a cron schedule. Currently defaults to `*/30 * * * * *`

### Logging Updates
- Updated logging for server to not log `OPTIONS` as it fills the whole log.

### Updates CORS & AUTH hook
- Updated the hook to `onRequest` to allow for request errors to be send on reply to UI


farhanW3:
@adam-maj 

Added `/backend-wallet/:chain/:walletAddress/get-nonces` 

- We return the DB nonce & not the on-chain nonce. If Wallet nonce is not found we return an error.

*Successful Response*

image

*Error Response*

image

farhanW3:
Added logic to chcek if the cron schedule provided is valid and within our constraints

PR #382, opened 2024-01-16
Simulate Tx : Request Option

farhanW3:
## Changes

Updated the implementation to allow devs to pass in `simulateTx` as a request ~param~ querystring to allow for transaction simulation before adding it to Queue


src/server/schemas/sharedApiSchemas.ts
```diff
@@ -24,6 +24,12 @@ export const contractParamSchema = Type.Object({
     examples: ["0xc8be6265C06aC376876b4F62670adB3c4d72EABA"],
     description: "Contract address on the chain",
   }),
+  simulateTx: Type.Optional(
```
adam-maj, line 27: This should be in querystring across everything, rather than URL

PR #384, opened 2024-01-23
fix: fetch chains from API to support newer chains

arcoraven:
## Changes

Since the chains package is a stale snapshot, newer chains are not discoverable by Engine.
Switching to the new async methods queries chains data from DB.

Performance impact should be minimal because the SDK caches chains by ID/slug forever after the initial request.

## How this PR will be tested

- [ ] Make a request to a newer chain like 80085.

### DIFF
--- a/src/server/utils/chain.ts
+++ b/src/server/utils/chain.ts
@@ -1,66 +1,54 @@
 import { Static } from "@sinclair/typebox";
-import { allChains, getChainByChainIdAsync } from "@thirdweb-dev/chains";
+import {
+  getChainByChainIdAsync,
+  getChainBySlugAsync,
+} from "@thirdweb-dev/chains";
 import { getConfig } from "../../utils/cache/getConfig";
 import { networkResponseSchema } from "../../utils/cache/getSdk";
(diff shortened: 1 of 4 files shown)


src/server/utils/chain.ts
```diff
@@ -1,66 +1,54 @@
...
-    if (config?.chainOverrides) {
       const parsedChainOverrides = JSON.parse(config.chainOverrides);
-      chainData = parsedChainOverrides.find(
-        (chainData: Static<typeof networkResponseSchema>) =>
-          chainData.slug === chain,
-      );
+      const chainData: Static<typeof networkResponseSchema> | undefined =
+        parsedChainOverrides.find(
+          (chainData: Static<typeof networkResponseSchema>) =>
+            chainData.slug === chain,
```
farhanW3, line 23: Sometimes, users can pass in `80001` instead of `mumbai`, this would not return the value. Can we also include the check to see if the `chainId` === `parseInt(chain)` if its a number. rest should be good.

also, we should look into lowercase both the input and the data being compared too, to handle `Polygon` vs `polygon`.

PR #386, opened 2024-01-30
add support for chainless domain relayer

IDubuque:
## Changes

- added support for chainless domain forwarders
- add 'chainid' in relayer interface to use chainless domain forwarders

## How this PR will be tested

- [x] should successfully relay transaction with chainless domain forwarder
- [ ] should not cause relayer regressions

## Output

### DIFF
--- a/src/server/routes/relayer/index.ts
+++ b/src/server/routes/relayer/index.ts
@@ -6,6 +6,7 @@ import {
   ERC20PermitAbi,
   ERC2771ContextAbi,
   ForwarderAbi,
+  ForwarderAbiEIP712ChainlessDomain,
   NativeMetaTransaction,
 } from "../../../constants/relayer";
 import { getRelayerById } from "../../../db/relayer/getRelayerById";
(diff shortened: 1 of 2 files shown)


src/server/routes/relayer/index.ts
```diff
@@ -30,6 +31,7 @@ const BodySchema = Type.Union([
       gas: Type.String(),
       nonce: Type.String(),
       data: Type.String(),
+      chainid: Type.Optional(Type.String()),
```
farhanW3, line 34: I guess `Chain-ID` has to be all lower case & not camel-case, as the SDK has `chainid` in the request body and the contract ABI verify function signature too has `chainid`?

IDubuque: Yeah, this was a bug caught during testing. It seems to match the evm block.chainid convention. I would like to make it camelcase, but think it would be confusing to users if we had cc interface, but lowercase when calling the contract. Also, they expect the request message params to be directly passed

PR #401, opened 2024-02-06
Optional ABI param for getContract

kumaryash90:
Provide option to pass ABI as a contract param. 

This will be helpful in cases where the contract is not verified or the ABI can't be resolved by the SDK for certain type of proxies. If the user has the ABI, can pass that as contract param.


src/server/schemas/sharedApiSchemas.ts
```diff
@@ -24,6 +24,11 @@ export const contractParamSchema = Type.Object({
     examples: ["0xc8be6265C06aC376876b4F62670adB3c4d72EABA"],
     description: "Contract address on the chain",
   }),
+  abi: Type.Optional(
+    Type.String({
```
farhanW3, line 28: should we make this Array? as I think SDK has Array everywhere

PR #402, opened 2024-02-07
chore: moved simulation to queueRawTx

gerrysaporito:
## Changes

- Linear: https://linear.app/thirdweb/issue/INF-171/
- Here are the changes made.

## How this PR will be tested

- [ ] Open the dashboard and click X. Result: A modal should appear.
- [ ] Call the /foo/bar API. Result: Returns 200 with "baz" in the response body.

## Output


farhanW3:
🚢

PR #403, opened 2024-02-08
support utf8 signMessage

IDubuque:
## Changes

- add support for signMessage utf8 bytes

## How this PR will be tested

- [x] should support utf8 bytes signMessage

### DIFF
--- a/src/server/routes/backend-wallet/signMessage.ts
+++ b/src/server/routes/backend-wallet/signMessage.ts
@@ -1,4 +1,5 @@
 import { Static, Type } from "@sinclair/typebox";
+import { ethers } from "ethers";
 import { FastifyInstance } from "fastify";
 import { StatusCodes } from "http-status-codes";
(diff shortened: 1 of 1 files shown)


src/server/routes/backend-wallet/signMessage.ts
```diff
@@ -7,6 +8,7 @@ import { walletAuthSchema } from "../../schemas/wallet";
 
 const BodySchema = Type.Object({
   message: Type.String(),
+  asBytes: Type.Optional(Type.Boolean()),
```
farhanW3, line 11: this maybe just me, `asBytes` makes me think, the signMessage is going to sign as bytes. maybe we can change it to `isBytes` which is clear to me, stating that the message is in bytes.

IDubuque: I was thinking sign string as bytes, but can change to isBytes.

PR #405, opened 2024-02-09
fix: revert simulations to only use the prepared tx

gerrysaporito:
## Changes

- Linear: https://linear.app/thirdweb/issue/INF-171/
- Here are the changes made.

## How this PR will be tested

- [ ] Open the dashboard and click X. Result: A modal should appear.
- [ ] Call the /foo/bar API. Result: Returns 200 with "baz" in the response body.

## Output

### DIFF
--- a/src/db/transactions/queueTx.ts
+++ b/src/db/transactions/queueTx.ts
@@ -7,6 +7,7 @@ import { BigNumber } from "ethers";
 import type { ContractExtension } from "../../schema/extension";
 import { PrismaTransaction } from "../../schema/prisma";
 import { queueTxRaw } from "./queueTxRaw";
+import { simulateTx } from "../../server/utils/simulateTx";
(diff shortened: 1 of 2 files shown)


src/db/transactions/queueTx.ts
```diff
@@ -41,6 +42,10 @@ export const queueTx = async ({
     value: BigNumber.from(await tx.getValue()).toHexString(),
   };
 
+  if (shouldSimulate) {
```
farhanW3, line 45: we also need to need to check `!deployedContractAddress` as we already simulate to get the address when deploying contract.

it was there earlier, [link](https://github.com/thirdweb-dev/engine/pull/402/files#diff-431cbd487d80223caf92d885eb093aaca1a2b370499fdac32064a69d6918edfdL64)

PR #425, opened 2024-02-23
chore: add 10% min on gas retries

arcoraven:
## Changes

- Adds a minimum of 10% when retrying gas


src/worker/tasks/updateMinedTx.ts
```diff
@@ -59,14 +60,19 @@ export const updateMinedTx = async () => {
...
-                    network: sdk.getProvider(),
-                  })
-                ).timestamp * 1000,
-              );
+              const block = await getBlock({
+                block: receipt.blockNumber,
+                network: sdk.getProvider(),
+              });
+              if (!block) {
+                throw `Block not found. txHash=${
```
arcoraven, line 68: Logging this until we find a repro case where the block returns without a valid timestamp.

PR #431, opened 2024-02-26
Feature: Contract Events

IDubuque:
## Changes

- add contract indexer functionality to engine

## How this PR will be tested

- [x] should be able to add a contract for indexing
- [x] should be able to remove a contract for indexing
- [x] should be able to get logs for the contract

### Subscribe to a contract (POST)
Subscribes to a contract and starts indexing
/contract/:chain/:contractAddress/events/subscribe
http://localhost:3005/contract/Polygon/0x3c499c542cef5e3811e1192ce70d8cc03d5c3359/events/subscribe

### Unsubscribe to a contract (POST)
Unsubscribes to a contract and stops indexing
/contract/:chain/:contractAddress/events/unsubscribe
http://localhost:3005/contract/Polygon/0x3c499c542cef5e3811e1192ce70d8cc03d5c3359/events/unsubscribe

### Get active subscriptions (GET)
Gets all the active contract subscriptions
/contract/events/get-subscriptions
http://localhost:3005/contract/events/get-subscriptions

### Get Contract Event Logs (GET)
Gets the event logs for a contract between fromBlock and toBlock
/contract/:chain/:contractAddress/events/get-logs?fromBlock=100&toBlock=200
http://localhost:3005/contract/Polygon/0x3c499c542cef5e3811e1192ce70d8cc03d5c3359/events/get-logs

### Get Indexed Block Range (GET)
Gets the range of blocks indexed for the subscribed contract
/contract/:chain/:contractAddress/events/get-indexed-blocks
http://localhost:3005/contract/Polygon/events/get-indexed-blocks

### Get Event Logs (GET)
Gets the event logs between blockTimestamp ranges or row creationTimestamps
contract/events/get-logs?fromBlockTimestamp=0&toBlockTimestamp=100
contract/events/get-logs?fromCreationTimestamp=0&toCreationTimestamp=100
http://localhost:3005/contract/events/get-logs

toCreationTimestamp is a requirement for pay.


farhanW3:
@IDubuque 

can we update the `PR` to include some details on how use the end-point with some same data/inputs. Just so can speed up the testing.

also, in your testing phase were you able to run this without issues in a multi-worker scenario?

src/db/chainIndexers/getChainIndexer.ts
```diff
@@ -0,0 +1,27 @@
...
+  const lastIndexedBlock = await prisma.$queryRaw<
+    { lastIndexedBlock: number }[]
+  >`
+    SELECT
+      "lastIndexedBlock"
+    FROM
+      "chain_indexers"
+    WHERE
+      "chainId"=${chainId}
+    FOR UPDATE NOWAIT
```
farhanW3, line 24: `FOR UPDATE NOWAIT` might give an issue/error, i.e., `Resource busy & cannot acquire ...` with multi-worker scenario

src/db/configuration/getConfiguration.ts
```diff
@@ -178,6 +178,8 @@ export const getConfiguration = async (): Promise<Config> => {
         minedTxListenerCronSchedule: "*/5 * * * * *",
         maxTxsToUpdate: 50,
         retryTxListenerCronSchedule: "*/30 * * * * *",
+        indexerListenerCronSchedule: "*/5 * * * * *",
+        maxBlocksToIndex: 5,
```
farhanW3, line 182: Do we want the users to toggle these values? if Yes, we need to add these new configs under the `configuration` end-points, and allow user to change this.

src/db/contractEventLogs/createContractEventLogs.ts
```diff
@@ -0,0 +1,35 @@
...
+  decodedLog?: any; // Assuming JSON object for decodedLog
+  eventName?: string;
+  timestamp: Date;
+  transactionIndex: number;
+  logIndex: number;
+}
+
+export interface BulkInsertContractLogsParams {
+  pgtx?: PrismaTransaction;
+  logs: ContractEventLogEntry[];
```
farhanW3, line 23: maybe we can use the Prisma Table Schema definition which should already be available and then we do not need to explicitly define this interface.

if we want to use this interface and mark it as `export` then we should look into adding into `sharedSchemas.ts` file.

src/db/contractSubscriptions/deleteContractSubscription.ts
```diff
@@ -0,0 +1,20 @@
...
+interface RemoveContractSubscriptionParams {
+  chainId: number;
+  contractAddress: string;
+}
+
+export const deleteContractSubscription = async ({
+  chainId,
+  contractAddress,
+}: RemoveContractSubscriptionParams) => {
+  return prisma.contractSubscriptions.delete({
```
farhanW3, line 12: should we delete the data from DB or keep a list of history & mark it as deleted on DB level? this, will help users to check what they have used before.

src/prisma/migrations/20240226232919_add_event_subscriptions/migration.sql
```diff
@@ -0,0 +1,54 @@
...
+    "data" TEXT NOT NULL,
+    "eventName" TEXT,
+    "decodedLog" JSONB,
+    "timestamp" TIMESTAMP(3) NOT NULL,
+    "transactionIndex" INTEGER NOT NULL,
+    "logIndex" INTEGER NOT NULL,
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updatedAt" TIMESTAMP(3) NOT NULL,
+
+    CONSTRAINT "contract_event_logs_pkey" PRIMARY KEY ("transactionHash","logIndex")
```
farhanW3, line 40: I think, we should look into adding other indexes too based on the query pattern which we are exposing/implementing to help with faster retrieval when we have huge dataset to filter from

src/prisma/migrations/20240226232919_add_event_subscriptions/migration.sql
```diff
@@ -0,0 +1,54 @@
+/*
+  Warnings:
+
+  - Added the required column `maxBlocksToIndex` to the `configuration` table without a default value. This is not possible if the table is not empty.
+
+*/
+-- AlterTable
+ALTER TABLE "configuration" ADD COLUMN     "indexerListenerCronSchedule" TEXT,
+ADD COLUMN     "maxBlocksToIndex" INTEGER NOT NULL;
```
farhanW3, line 9: this will need a `DEFAULT` value if you mark it as `NOT NULL`.

Either mark it ass `nullable` or add a `default` value.

otherwise, get this error:

```
Database error:
ERROR: column "maxBlocksToIndex" of relation "configuration" contains null values

DbError { severity: "ERROR", parsed_severity: Some(Error), code: SqlState(E23502), message: "column \"maxBlocksToIndex\" of relation \"configuration\" contains null values", detail: None, hint: None, position: None, where_: None, schema: Some("public"), table: Some("configuration"), column: Some("maxBlocksToIndex"), datatype: None, constraint: None, file: Some("tablecmds.c"), line: Some(6046), routine: Some("ATRewriteTable") }
```

src/prisma/schema.prisma
```diff
@@ -23,6 +23,9 @@ model Configuration {
   maxFeePerGasForRetries             String  @map("maxFeePerGasForRetries")
   maxPriorityFeePerGasForRetries     String  @map("maxPriorityFeePerGasForRetries")
   maxRetriesPerTx                    Int     @map("maxRetriesPerTx")
+  // Contract Indexer Updates
+  indexerListenerCronSchedule        String? @map("indexerListenerCronSchedule")
+  maxBlocksToIndex                   Int     @map("maxBlocksToIndex")
```
farhanW3, line 28: `maxBlocksToIndex` :   will need a DEFAULT value if you mark it as NOT NULL.

Either mark it as nullable and on code level provide the value which you currently do (i.e., 5 i believe) or add a default value.

otherwise, get this error, when running the server :

```
Database error:
ERROR: column "maxBlocksToIndex" of relation "configuration" contains null values

DbError { severity: "ERROR", parsed_severity: Some(Error), code: SqlState(E23502), message: "column \"maxBlocksToIndex\" of relation \"configuration\" contains null values", detail: None, hint: None, position: None, where_: None, schema: Some("public"), table: Some("configuration"), column: Some("maxBlocksToIndex"), datatype: None, constraint: None, file: Some("tablecmds.c"), line: Some(6046), routine: Some("ATRewriteTable") }
```

src/server/routes/contract/events/addContractSubscription.ts
```diff
@@ -0,0 +1,90 @@
...
+        }
+      }
+
+      // upsert indexed contract, this will be picked up
+      await upsertContractSubscription({
+        chainId,
+        contractAddress: standardizedContractAddress,
+      });
+
+      logger({
```
farhanW3, line 75: don't think, we need these logs.

src/server/routes/contract/events/getContractEventLogs.ts
```diff
@@ -0,0 +1,132 @@
...
+      params: contractParamSchema,
+      querystring: requestQuerySchema,
+      response: {
+        ...standardResponseSchema,
+        [StatusCodes.OK]: responseSchema,
+      },
+    },
+    handler: async (request, reply) => {
+      const { chain, contractAddress } = request.params;
+      const { fromBlock, toBlock, topics } = request.query;
```
farhanW3, line 72: we should also maybe have a limit on the number of blocks user can query at once, `toBlock - fromBlock <= 100` or something.

src/db/chainIndexers/getChainIndexer.ts
IDubuque: This is expected, we should just catch the error

src/utils/cache/getSdk.ts
```diff
@@ -77,7 +77,9 @@ export const getSdk = async ({
   let chain: NetworkInput | undefined = undefined;
   try {
     chain = await getChainByChainIdAsync(chainId);
-  } catch (error) {}
+  } catch (error) {
+    console.log("failed to get the chain");
```
farhanW3, line 81: can you use `logger()` from our utils folder, instead of `console.log()`. and do we need this error log? is it actionable?

src/server/routes/contract/events/addContractSubscription.ts
```diff
@@ -0,0 +1,90 @@
...
+      tags: ["Contract", "Index"],
+      operationId: "write",
+      params: contractParamSchema,
+      response: {
+        ...standardResponseSchema,
+        [StatusCodes.OK]: responseSchema,
+      },
+    },
+    handler: async (request, reply) => {
+      const { chain, contractAddress } = request.params;
```
farhanW3, line 50: are we supporting all chains?? if `Yes` then ignore, if not then we need to have an allowList

src/db/contractEventLogs/createContractEventLogs.ts
IDubuque: makes sense. Can change to use generated schema

src/db/contractSubscriptions/deleteContractSubscription.ts
IDubuque: we can mark inactive with deletedAt field

src/worker/indexers/chainIndexer.ts
```diff
@@ -0,0 +1,285 @@
...
+  const sdk = await getSdk({ chainId: params.chainId });
+  const provider = sdk.getProvider();
+
+  logger({
+    service: "worker",
+    level: "debug",
+    message: `Starting log fetch - ChainIndexer: ${params.chainId}`,
+  });
+
+  const startTime = performance.now();
```
farhanW3, line 40: we should remove these as we do need this in main. good when developing as we want to see how much time it takes etc, but not required for `main`

src/utils/cache/getSdk.ts
IDubuque: yeah, this was a debug log that was missed

src/server/routes/contract/events/addContractSubscription.ts
IDubuque: all chains

src/worker/indexers/chainIndexer.ts
```diff
@@ -0,0 +1,285 @@
...
+  const contracts = await Promise.all(
+    uniqueContractAddresses.map(async (address) => {
+      const contract = await getContract({
+        chainId: params.chainId,
+        contractAddress: address,
+      });
+      return { contractAddress: address, contract: contract };
+    }),
+  );
+  const contractCache = contracts.reduce((acc, val) => {
```
farhanW3, line 96: need to be careful with caching, as a `partial-ABI` can lead to issues. Also, contract ABI/Metadata is cached on by our infra service, namely `contract-api`.

src/worker/indexers/chainIndexer.ts
```diff
@@ -0,0 +1,285 @@
...
+      timestamp: new Date(block.timestamp * 1000), // ethers timestamp is s, Date uses ms
+      transactionIndex: log.transactionIndex,
+      logIndex: log.logIndex,
+    } as ContractEventLogEntry;
+  });
+
+  return formattedLogs;
+};
+
+export const createChainIndexerTask = async (chainId: number) => {
```
farhanW3, line 192: can we move this into the `tasks` folder with a new file. to be consistent on the structure of files/folders

src/worker/tasks/manageChainIndexers.ts
```diff
@@ -0,0 +1,23 @@
+import { getContractSubscriptionsUniqueChainIds } from "../../db/contractSubscriptions/getContractSubscriptions";
+import {
+  INDEXER_REGISTRY,
+  addChainIndexer,
+  removeChainIndexer,
+} from "../indexers/chainIndexerRegistry";
+
+export const manageChainIndexers = async () => {
```
farhanW3, line 8: this step here:

- it's creating cron-jobs and then clearing the created cron-jobs??

src/db/chainIndexers/getChainIndexer.ts
farhanW3: got it

src/db/configuration/getConfiguration.ts
farhanW3: maybe this was not clear,

we need to add `/configuration/indexer` GET & POST and include the above properties.

src/worker/indexers/chainIndexer.ts
IDubuque: this is just locally caching the result of the calls. I guess this line isn't needed, but good to know we cache in sdk. Above should be fine tho

src/worker/tasks/manageChainIndexers.ts
IDubuque: yup. If indexer no longer needed, we take down the cron job. No reason to run a job if chain is no longer indexed

src/server/routes/contract/events/getEventLogs.ts
```diff
@@ -0,0 +1,159 @@
...
+        toBlockTimestamp,
+      } = request.query;
+
+      const hasCreatedTimestamps =
+        fromCreationTimestamp !== undefined ||
+        toCreationTimestamp !== undefined;
+      const hasBlockTimestamps =
+        fromBlockTimestamp !== undefined || toBlockTimestamp !== undefined;
+
+      if (hasCreatedTimestamps == hasBlockTimestamps) {
```
farhanW3, line 83: @IDubuque 

we should update the `requestQuerySchema` to have the fields as required and let `typebox` do the checks, to be consistent. If needed, we can handle the `typbox` error message much better and display a more dev friendly message.

farhanW3:
Found few other things:

1. The field: `indexerListenerCronSchedule` since is set to `nullable` is NULL on my config table, thus the cron doesn't get scheduled.

To handle this, you will need to update the `getConfiguration()` implementation to include the below

```
...
else if (!config.authDomain && !config.authWalletEncryptedJson) {
    // TODO: Use a more generic method to fill missing fields
    config = await updateConfiguration({
      authDomain: "thirdweb.com",
      authWalletEncryptedJson: await createAuthWalletEncryptedJson(),
    });
  } else if (!config.accessControlAllowOrigin) {
    config = await updateConfiguration({
      accessControlAllowOrigin: mandatoryAllowedCorsUrls.join(","),
    });
  } else if (!config.indexerListenerCronSchedule) { // New contract indexer cron update
    config = await updateConfiguration({
      indexerListenerCronSchedule: "*/5 * * * * *", // maybe
    });
}
```

2. Add new indexer routes under `configuration` folder to get & update the `indexer` config variables. Similar to have we have for others

src/utils/indexer/getBlockTime.ts
```diff
@@ -0,0 +1,44 @@
...
+    for (let i = 0; i < BLOCKS_TO_ESTIMATE; i++) {
+      const block = await provider.getBlock(latestBlockNumber - 1);
+      blocks.push(block);
+    }
+
+    let totalTimeDiff = 0;
+    for (let i = 0; i < blocks.length - 1; i++) {
+      totalTimeDiff += blocks[i].timestamp - blocks[i + 1].timestamp;
+    }
+    const averageBlockTime = totalTimeDiff / (blocks.length - 1);
```
farhanW3, line 32: @IDubuque 

this needs to be `i`. to get everything working.

farhanW3: also, I m not able to get any Events for my [contract](https://mumbai.polygonscan.com/address/0x365b83d67d5539c6583b9c0266a548926bf216f4) on `mumbai`.

you you try and change the DB value for `lastIndexedBlock` to `46436824` and you should be seeings logs/events coming in and getting added to DB. but I don't see anything there

IDubuque: Nice catch, missed this typo and only tested with hardcoded chains

src/server/routes/contract/events/getContractIndexedBlockRange.ts
```diff
@@ -0,0 +1,81 @@
...
+    },
+    handler: async (request, reply) => {
+      const { chain, contractAddress } = request.params;
+      const standardizedContractAddress = contractAddress.toLowerCase();
+
+      const chainId = await getChainIdFromChain(chain);
+
+      const result = await getContractEventLogsIndexedBlockRange({
+        chainId,
+        contractAddress,
```
farhanW3, line 58: this should be `standardizedContractAddress` instead of `contractAddress`

src/server/routes/contract/events/getEventLogs.ts
```diff
@@ -0,0 +1,142 @@
...
+  Type.Object({
+    contractAddresses: Type.Optional(Type.Array(Type.String())),
+    topics: Type.Optional(Type.Array(Type.String())),
+    fromCreationTimestamp: Type.Number(),
+    toCreationTimestamp: Type.Optional(Type.Number()),
+  }),
+  Type.Object({
+    contractAddresses: Type.Optional(Type.Array(Type.String())),
+    topics: Type.Optional(Type.Array(Type.String())),
+    fromBlockTimestamp: Type.Number(),
```
farhanW3, line 21: should this be `fromBlockNumber`??

IDubuque: No, block number is chain dependent. If I want to get all the events on my contracts across different chains that happened within a certain period of time, I need fromBlockTimestamp, toBlockTimestamp. 

Also, if I am querying a moving window in my application:
1. set lastIndexedTimestamp = new Date.getTime()
2. set currentTimestamp = new Date.getTime()
3. query { lastIndexedTimestamp, currentTimestamp }
4. set lastIndexedTimestamp = currentTimestamp
5. jmp 2

then I cannot use fromBlockTimestamp, I need to use fromCreationTimestamp. This is because the indexer will have a slight delay (that will vary by workload), causing the indexer to add blockTimestamp at blockTimestamp + delay. Using creation time will negate the need for adding a look back to the sliding window especially bc it could vary.

farhanW3: > If I want to get all the events on my contracts across different chains that happened within a certain period of time, I need fromBlockTimestamp, toBlockTimestamp.

Then can't we achieve the same with `fromCreationTimestamp`/`toCreationTimestamp`??

src/db/contractEventLogs/getContractEventLogs.ts
```diff
@@ -0,0 +1,203 @@
...
+                  blockNumber: { equals: cursor.blockNumber },
+                  transactionIndex: { gt: cursor.transactionIndex },
+                },
+                {
+                  createdAt: { equals: cursor.createdAt },
+                  chainId: { equals: cursor.chainId },
+                  blockNumber: { equals: cursor.blockNumber },
+                  transactionIndex: { equals: cursor.transactionIndex },
+                  logIndex: { gt: cursor.logIndex },
+                },
```
farhanW3, line 135: Do we need a composite index for this as it can slow down retrieval with huge data volume.

src/server/routes/contract/transactions/getTransactionReceipts.ts
```diff
@@ -0,0 +1,148 @@
...
+          timestamp: txRcpt.timestamp.getTime(),
+          data: txRcpt.data,
+          value: txRcpt.value,
+          to: txRcpt.to,
+          from: txRcpt.from,
+          transactionIndex: txRcpt.transactionIndex,
+
+          gasUsed: txRcpt.gasUsed,
+          effectiveGasPrice: txRcpt.effectiveGasPrice,
+          status: txRcpt.status,
```
farhanW3, line 136: nit, we can do {...txRcpt, timestamp: txRcpt.timestamp.getTime()} i feel.

src/server/routes/contract/transactions/getTransactionReceipts.ts
```diff
@@ -0,0 +1,148 @@
...
+        };
+      });
+
+      reply.status(StatusCodes.OK).send({
+        result: {
+          receipts: transactionReceipts,
+          status: "success",
```
farhanW3, line 143: we don't really sent a status in our other end-points. we can remove this

src/server/routes/contract/events/paginateEventLogs.ts
```diff
@@ -0,0 +1,193 @@
...
+
+          cursorObj = validationResult.data;
+        }
+      } catch (error) {
+        throw createCustomError(
+          "Invalid cursor",
+          StatusCodes.BAD_REQUEST,
+          "BAD_REQUEST",
+        );
+      }
```
farhanW3, line 134: can we put this logic inside getTransactionReceiptsByCursor() so that others using this function can use the logic

src/worker/tasks/chainIndexer.ts
```diff
@@ -0,0 +1,356 @@
...
+
+  const txReceipts = transactionsWithReceipt.map(({ receipt, transaction }) => {
+    return {
+      chainId: chainId,
+      blockNumber: receipt.blockNumber,
+      contractAddress: receipt.to.toLowerCase(),
+      contractId: getContractId(chainId, receipt.to.toLowerCase()),
+      transactionHash: receipt.transactionHash.toLowerCase(),
+      blockHash: receipt.blockHash.toLowerCase(),
+      timestamp: new Date(blockLookup[receipt.blockNumber].timestamp),
```
farhanW3, line 244: this needs to be updated.

Looks like all dates are coming incorrect, `1970-01-20 19:06:37.024`. 

We need to multiply by `1000` as the timestamp is in Unix Epoc (seconds

PR #432, opened 2024-02-26
feat: show clearer message if out of gas

arcoraven:
## Changes

Maps the out of gas error to a clearer error message:

Old:
```
insufficient funds for intrinsic transaction cost [ See: https://links.ethers.org/v5-errors-INSUFFICIENT_FUNDS ] (error={"code":3}, method="estimateGas", transaction={"from":"0x4Ff9aa707AE1eAeb40E581DF2cf4e14AffcC553d","maxPriorityFeePerGas":{"type":"BigNumber","hex":"0x59682f00"},"maxFeePerGas":{"type":"BigNumber","hex":"0x59af3c76"},"to":"0x4Ff9aa707AE1eAeb40E581DF2cf4e14AffcC553d","data":"0x","type":2,"accessList":null}, code=INSUFFICIENT_FUNDS, version=providers/5.7.2)
 code: "INSUFFICIENT_FUNDS"
```

New:
```
Insufficient ETH in backend wallet 0x4Ff9aa707AE1eAeb40E581DF2cf4e14AffcC553d.
```


src/db/transactions/getQueuedTxs.ts
```diff
@@ -35,5 +32,5 @@ export const getQueuedTxs = async ({ pgtx }: GetQueuedTxsParams = {}): Promise<
   FOR UPDATE SKIP LOCKED
   `;
 
-  return cleanTxs(txs);
```
arcoraven, line 38: Removing an unnecessary abstraction

src/utils/errors.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const parseTxError = async (
+  tx: Transactions,
+  err: any,
+): Promise<string | null> => {
+  if (!err) {
+    return null;
+  }
+
+  if ((err as EthersError)?.code === ethers.errors.INSUFFICIENT_FUNDS) {
```
arcoraven, line 21: Basically maps the insufficient funds error code to a more human-readable error message.

src/worker/tasks/processTx.ts
```diff
@@ -152,17 +149,16 @@ export const processTx = async () => {
           }
         }
 
-        // 3. Group transactions to be batched by sender address & chain id
-        const txsByWallet = txsToSend.reduce((acc, curr) => {
-          const key = `${curr.fromAddress}-${curr.chainId}`;
-          if (key in acc) {
-            acc[key].push(curr);
+        // 3. Group transactions to be batched by sender address and chain.
+        const txsByWallet: Record<string, Transactions[]> = {};
```
arcoraven, line 153: Rewriting reduce to a simpler for loop

src/worker/tasks/processTx.ts
```diff
@@ -230,14 +226,6 @@ export const processTx = async () => {
               });
             }
 
-            if (!dbNonceData) {
```
arcoraven, line 233: This isn't a real "error" anymore since we're treating the db nonce as a cache.

src/utils/usage.ts
```diff
@@ -29,7 +29,6 @@ export interface ReportUsageParams {
     extension?: string;
     retryCount?: number;
     provider?: string;
-    transactionValue?: string;
```
arcoraven, line 32: Unused

src/worker/tasks/processTx.ts
```diff
@@ -355,16 +295,12 @@ export const processTx = async () => {
                   txIndex++;
 
                   rpcResponses.push({
-                    queueId: tx.queueId!,
-                    tx: txRequest,
-                    res: rpcResponse,
-                    sentAt: new Date(),
-                    functionName: tx.functionName || undefined,
-                    extension: tx.extension || undefined,
-                    queuedAt: new Date(tx.queuedAt!).getTime(),
```
arcoraven, line 364: Most of these fields are from `tx`, the DB transaction object. I'm adding tx to the payload to simplify readability. Now we're passing around `(tx db object, ethers tx request, rpc response)`.

src/worker/tasks/processTx.ts
```diff
@@ -152,469 +108,279 @@ export const processTx = async () => {
...
-                        provider: provider.connection.url || undefined,
-                        msSinceQueue: new Date().getTime() - tx.queuedAt!,
-                      },
-                      action: UsageEventTxActionEnum.NotSendTx,
-                    });
-                    break;
-                }
-              }),
-            );
-          } catch (err: any) {
```
arcoraven, line 571: Removing this `catch` block. I don't think we should error out txs if this code block throws because it's likely unrelated issues (RPC, wallet).

farhanW3: this try/catch helps in catching these:

```
QUEUED:INCREASE_ALLOWANCE:ENGINE:ENGINE_FAILURE:[Worker] [Error] Failed to process batch of transactions for wallet - Error: timeout (requestBody="[{\"method\":\"eth_blockNumber\",\"params\":[],\"id\":5521,\"jsonrpc\":\"2.0\"}]", requestMethod="POST", timeout=120000, url="https://polygon.rpc.thirdweb.com/fc01ea4fbd46234e23098c7daa7b6ef2", code=TIMEOUT, version=web/5.7.1)
```

this line, i believe, `sentAtBlockNumber: await provider.getBlockNumber(),` is getting caught there.

similarly, other RPC calls to get tx details etc gets caught here. maybe we need to update this logic

Maybe we can add a try-catch inside the match and handle the updatesToDB when transasction was processed but couldn't get data via RPC more gracefully, instead of just marking them as errored.

src/utils/errors.ts
farhanW3: I logged `err` :

```
{
  code: 3,
  message: 'insufficient funds for gas * price + value: balance 1636106304978556, queued cost 1570206001522624, tx cost 1570206001522624, overshot 1504305698066692'
}
```

and looks like `err.code` is `3` when the `insufficient ...` message comes from RPC and `ethers.errors.INSUFFICIENT_FUNDS` is `INSUFFICIENT_FUNDS`, so they check fails. though the RPC message is elaborate but the balance value seems off.

also, when wallet-balance is low, gas-estimation fails too:

```Error: cannot estimate gas; transaction may fail or may require manual gas limit [ See: https://links.ethers.org/v5-errors-UNPREDICTABLE_GAS_LIMIT ] (error={"reason":"cannot estimate gas; transaction may fail or may require manual gas limit","code":"UNPREDICTABLE_GAS_LIMIT","error":{"code":3},"method":"estimateGas","transaction":{"from":"0x3EcDBF3B911d0e9052b64850693888b008e18373","maxPriorityFeePerGas":{"type":"BigNumber","hex":"0x07aef40a00"},"maxFeePerGas":{"type":"BigNumber","hex":"0x07aef40a20"},"to":"0x365b83D67D5539C6583b9c0266A548926Bf216F4","value":{"type":"BigNumber","hex":"0x00"},"data":"0xa9059cbb0000000000000000000000003ecdbf3b911d0e9052b64850693888b008e1837300000000000000000000000000000000000000000000000000766c7d74835555","type":2,"accessList":null}}, tx={"to":{},"from":"0x3ecdbf3b911d0e9052b64850693888b008e18373","data":"0xa9059cbb0000000000000000000000003ecdbf3b911d0e9052b64850693888b008e1837300000000000000000000000000000000000000000000000000766c7d74835555","value":{"type":"BigNumber","hex":"0x00"},"nonce":{"type":"BigNumber","hex":"0x1836"},"maxFeePerGas":{"type":"BigNumber","hex":"0x07aef40a20"},"maxPriorityFeePerGas":{"type":"BigNumber","hex":"0x07aef40a00"},"type":2,"gasLimit":{},"chainId":{}}, code=UNPREDICTABLE_GAS_LIMIT, version=abstract-signer/5.7.0)```

PR #442, opened 2024-03-05
fix: remove gas limit on retries

arcoraven:
- Remove gas limit on retries. This was causing txns to get stuck during moderate gas spikes. We will revisit per-tx gas limits in the future.
- Make retry logic more aggressive (i.e. faster).
- Properly cancel a tx if it times out.
- Better intro message to instruct users to dashboard.


src/worker/tasks/updateMinedTx.ts
```diff
@@ -144,45 +168,6 @@ export const updateMinedTx = async () => {
             });
           }),
         );
-
-        // Update dropped txs.
-        await Promise.all(
```
arcoraven, line 149: No longer used because we're cancelling txs synchronously in this flow.

PR #446, opened 2024-03-13
Feature/contract tx indexer

IDubuque:
## Changes

- add transaction receipts for subscribed contracts

## How this PR will be tested

- [x] confirmed transaction receipts are being tracked locally

## Output

### DIFF
--- a/src/db/contractTransactionReceipts/getContractTransactionReceipts.ts
+++ b/src/db/contractTransactionReceipts/getContractTransactionReceipts.ts
@@ -51,7 +51,7 @@ export const getTransactionReceiptsByBlockTimestamp = async ({
       : {}),
   };
 
-  return await prisma.contractEventLogs.findMany({
+  return await prisma.contractTransactionReceipts.findMany({
     where: whereClause,
   });
 };
--- a/src/db/contractTransactionReceipts/getContractTransactionReceipts.ts
+++ b/src/db/contractTransactionReceipts/getContractTransactionReceipts.ts
@@ -1,0 +1,130 @@
+import { prisma } from "../client";
+
+interface GetContractTransactionReceiptsParams {
+  chainId: number;
+  contractAddress: string;
+  fromBlock: number;
+  toBlock?: number;
+}
+
+export const getContractTransactionReceiptsByBlock = async ({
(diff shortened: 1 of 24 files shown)


src/db/contractTransactionReceipts/getContractTransactionReceipts.ts
```diff
@@ -0,0 +1,130 @@
...
+                  createdAt: { equals: cursor.createdAt },
+                  chainId: { equals: cursor.chainId },
+                  blockNumber: { gt: cursor.blockNumber },
+                },
+                {
+                  createdAt: { equals: cursor.createdAt },
+                  chainId: { equals: cursor.chainId },
+                  blockNumber: { gt: cursor.blockNumber },
+                  transactionIndex: { gt: cursor.transactionIndex },
+                },
```
farhanW3, line 101: Do we need a composite index for this as it can slow down retrieval with huge data volume.

src/server/routes/contract/transactions/getTransactionReceiptsByTimestamp.ts
```diff
@@ -0,0 +1,110 @@
...
+          gasUsed: txRcpt.gasUsed,
+          effectiveGasPrice: txRcpt.effectiveGasPrice,
+          status: txRcpt.status,
+        };
+      });
+
+      reply.status(StatusCodes.OK).send({
+        result: {
+          receipts: transactionReceipts,
+          status: "success",
```
farhanW3, line 105: we don't really sent a `status` in our other end-points. we can remove this

src/server/routes/contract/transactions/paginateTransactionReceipts.ts
```diff
@@ -0,0 +1,176 @@
...
+          cursorObj = validationResult.data;
+        }
+      } catch (error) {
+        throw createCustomError(
+          "Invalid cursor",
+          StatusCodes.BAD_REQUEST,
+          "BAD_REQUEST",
+        );
+      }
+
```
farhanW3, line 124: can we put this logic inside `getTransactionReceiptsByCursor()` so that others using this function can use the logic

PR #454, opened 2024-03-15
chore: Clean up TransactionStatus enum

arcoraven:
## Changes

- Removes "retried" and "processed" statuses. They are unused.
- Removed related webhooks.


src/worker/tasks/deleteTx.ts
```diff
@@ -8,12 +8,13 @@ export const deleteTx = async () => {
...
     const deletedItems = await prisma.transactions.deleteMany({
       where: {
+        // All txs queued 24+ hours ago that are mined, cancelled, or errored.
         AND: [
           {
             queuedAt: {
               lt: twentyFourHoursAgo,
             },
-            processedAt: {
+            sentAt: {
```
arcoraven, line 17: remove this

src/db/transactions/getQueueStatus.ts
```diff
@@ -10,7 +10,7 @@ export const getQueueStatus = async ({
   const queued = await prisma.transactions.count({
     where: {
       fromAddress: walletAddress?.toLowerCase(),
-      processedAt: null,
+      sentAt: null,
       errorMessage: null,
     },
```
arcoraven, line 15:
```
sentAt: null
errorMessage: null
cancelledAt: null
```

src/worker/tasks/processTx.ts
```diff
@@ -72,14 +72,6 @@ export const processTx = async () => {
             message: `Processing`,
           });
 
-          await updateTx({
```
arcoraven, line 75: This is a big win to remove a unnecessary DB write on each tx request.

PR #455, opened 2024-03-16
feat: Add webhook passthrough params

arcoraven:
## Changes

- Adds field `customMetadata` to all write requests as a passthrough string that will be set on all webhooks.


src/server/schemas/transaction/index.ts
```diff
@@ -201,41 +202,3 @@ export enum TransactionStatusEnum {
   Cancelled = "cancelled",
   Retried = "retried",
 }
-
-export interface TransactionSchema {
```
arcoraven, line 205: Unused

PR #459, opened 2024-03-19
change cursor delay to ms

IDubuque:
## Changes

- change cursor delay from s to ms


farhanW3:
you will need a migration file generated for this SQL update to take place. @IDubuque

PR #461, opened 2024-03-20
fix: Updated minedTx flow + removed deleteAllNonce on start-up

farhanW3:
Updated minedTx Flow:
 - Added try-catch to handle error on `cancelTransactionAndUpdate()`. On Error, mark the transaction as Errored.

Removed `deleteAllNonce()` on start-up:
 - This was causing nonce desync on restart as it was deleting nonce table information.

### DIFF
--- a/src/worker/tasks/updateMinedTx.ts
+++ b/src/worker/tasks/updateMinedTx.ts
@@ -48,28 +48,57 @@ export const updateMinedTx = async () => {
                 const sentAt = new Date(tx.sentAt!);
                 const ageInMilliseconds = Date.now() - sentAt.getTime();
                 if (ageInMilliseconds > MEMPOOL_DURATION_TIMEOUT_MS) {
-                  await cancelTransactionAndUpdate({
-                    queueId: tx.id,
-                    pgtx,
-                  });
+                  try {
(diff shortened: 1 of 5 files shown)


src/worker/tasks/updateMinedTx.ts
```diff
@@ -48,28 +48,57 @@ export const updateMinedTx = async () => {
...
+                      input: {
+                        fromAddress: tx.fromAddress || undefined,
+                        toAddress: tx.toAddress || undefined,
+                        value: tx.value || undefined,
+                        chainId: tx.chainId || undefined,
+                        transactionHash: tx.transactionHash || undefined,
+                        provider: provider.connection.url || undefined,
+                        msSinceSend: Date.now() - tx.sentAt!.getTime(),
+                      },
+                      action: UsageEventTxActionEnum.NotSendTx,
```
arcoraven, line 99: We should report these as "errors" (Engine problem to fix) instead of "not send" (customer problem to fix)

farhanW3: we do report `notsendTx` in other places too, like the retryTx Flow also i believe should follow same notion. I will update

PR #463, opened 2024-03-21
feat: Add idempotency key support

arcoraven:
## Changes

- Adds support for an `x-idempotency-key` header.
- This allows a tx to be retried multiple times and ensure that it will only be enqueued once.


src/db/transactions/queueTxRaw.ts
```diff
@@ -48,21 +52,42 @@ export const queueTxRaw = async ({
...
-    data: {
-      ...tx,
-      fromAddress: tx.fromAddress?.toLowerCase(),
-      toAddress: tx.toAddress?.toLowerCase(),
-      target: tx.target?.toLowerCase(),
-      signerAddress: tx.signerAddress?.toLowerCase(),
-      accountAddress: tx.accountAddress?.toLowerCase(),
-    },
-  });
+  const insertData = {
```
arcoraven, line 55: This is the main change:
```
If idempotencyKey is provided, upsert
  If the row exists
    Return the existing row (does not enqueue a new tx)
  If the row does not exist
    Insert the row (enqueue a new tx)
Else
  Insert the row (enqueue a new tx)

File: src/prisma/schema.prisma [hunk 2]
```diff
@@ -103,6 +103,7 @@ model WalletNonce {
 
 model Transactions {
   id                        String    @id @default(uuid()) @map("id")
+  idempotencyKey            String    @unique @default(uuid()) @map("idempotencyKey")
```
arcoraven, line 106: - `@unique` - Creates a unique index for O(1) lookups. Ensures exactly one tx for a given idempotency key.
- `@default(uuid())` - For backward compatibility. When introducing this change to a new DB, there needs to be non-null values to satisfy the uniqueness constraint.

src/server/schemas/txOverrides.ts
```diff
@@ -0,0 +1,21 @@
+import { Type } from "@sinclair/typebox";
```
arcoraven, line 1: Just renamed this file (`web3api-overrides/index.ts` -> `txOverrides.ts`)

PR #477, opened 2024-04-03
add gas to txOverrides

joaquim-verges:
## PR-Codex overview
This PR introduces gas-related enhancements to transaction processing, including gas limit overrides and gas price adjustments.

### Detailed summary
- Added `gas` field to transaction overrides
- Included `gas`, `maxFeePerGas`, and `maxPriorityFeePerGas` in transaction schema
- Updated transaction preparation in `write.ts` to include gas-related overrides
- Modified `queueTx` function to accept `txOverrides` with gas-related fields
- Implemented gas limit and gas price overrides in `processTx.ts`


arcoraven (approved):

farhanW3:
This PR needs more work.

We only updated the Ingest to take in the `gas` value on `txOverride` input, but we also need to make sure to store the gas-values on DB when queueing, and update the `processTx` flow to check if the gas-values fields are populated on DB and use it for the transaction otherwise use `getDefaultGasOverrides()`.

@arcoraven @joaquim-verges

0xFirekeeper (approved):

src/worker/tasks/processTx.ts
```diff
@@ -378,6 +394,24 @@ export const processTx = async () => {
...
+              const paymasterResult =
+                await signer.smartAccountAPI.paymasterAPI.getPaymasterAndData(
+                  unsignedOp,
+                );
+              const paymasterAndData = paymasterResult.paymasterAndData;
+              if (paymasterAndData && paymasterAndData !== "0x") {
+                unsignedOp.paymasterAndData = paymasterAndData;
+              }
+            }
+
```
farhanW3, line 414: this is to help set custom `callGasLimit` for userOp.

src/worker/tasks/processTx.ts
```diff
@@ -171,11 +171,27 @@ export const processTx = async () => {
...
+              // Gas limit override
+              if (tx.gasLimit) {
+                txRequest.gasLimit = BigNumber.from(tx.gasLimit);
+              } else {
+                // TODO: We need to target specific cases
+                // Bump gas limit to avoid occasional out of gas errors
+                txRequest.gasLimit = txRequest.gasLimit
+                  ? BigNumber.from(txRequest.gasLimit).mul(120).div(100)
+                  : undefined;
+              }
```
arcoraven, line 183:
```suggestion
} else if (txRequest.gasLimit) {
  // Use +20% gas limit to prevent occasional gas errors.
  txRequest.gasLimit = BigNumber.from(txRequest.gasLimit).mul(120).div(100)
}
```

0xFirekeeper: If they pass a gas limit we should probably not modify it, default without gas limit overrides already adds overheads

arcoraven: I believe this was added for shield (ran into issues running out of gas on airdrops). We could remove it and see if it returns?

arcoraven (approved):

PR #483, opened 2024-04-08
Ph/send worker

arcoraven:
@TODO


src/worker/tasks/queuedTxWorker.ts
```diff
@@ -0,0 +1,248 @@
...
+        address: fromAddress,
+        client: thirdwebClient,
+        chain,
+      });
+
+      // Estimate gas for a transfer.
+      const transferTx = prepareTransaction({
+        client: thirdwebClient,
+        chain,
+        value: BigInt(1),
```
farhanW3, line 209: this should be `balanceWei` i believe & not `1`.

src/worker/tasks/sentTxWorker.ts
```diff
@@ -0,0 +1,253 @@
...
+
+    // Update the transactions DB.
+    await updateTx({
+      queueId: sentTransaction.id,
+      data: {
+        status: TransactionStatus.Sent,
+        sentAt: new Date(),
+        sentAtBlock: await eth_blockNumber(rpcRequest),
+        transactionHash,
+        retryCount: 1,
```
farhanW3, line 170: we should do `dbData.retryCount + 1` or update `updateTx()` to increment on `Retry` if not here.

PR #494, opened 2024-04-11
feat(experimental): Keypair auth

arcoraven:
## PR-Codex overview
This PR focuses on adding keypair functionality and improving caching. 

### Detailed summary
- Added `Keypairs` model with label and algorithm fields
- Implemented keypair CRUD operations
- Updated encryption functions
- Enhanced keypair caching
- Added keypair-related routes for listing and removing public keys
- Improved health check to include keypair auth feature
- Updated environment configuration for testing

> The following files were skipped due to too many changes: `src/utils/env.ts`, `src/server/schemas/keypairs.ts`, `src/server/routes/auth/keypair/add.ts`, `yarn.lock`, `src/tests/auth.test.ts`, `src/server/middleware/auth.ts`


src/server/routes/auth/keypair/revoke.ts
```diff
@@ -0,0 +1,49 @@
...
+        ...standardResponseSchema,
+        [StatusCodes.OK]: ReplySchema,
+      },
+    },
+    handler: async (req, res) => {
+      const { hash } = req.body;
+
+      await deleteKeypair({ hash });
+      keypairCache.clear();
+
```
farhanW3, line 41: Do we want to check the DB before deletion? else, always end up sending `OK`. We do the same on webhooks revoke

src/server/routes/auth/keypair/revoke.ts
```diff
@@ -0,0 +1,49 @@
...
+      res.status(200).send({
+        result: {
+          message: "OK",
+        },
```
farhanW3, line 45: We respond with `success: true` on other end-points. So keeping it consistent would be great

```
res.status(200).send({
        result: {
          success: true,
        },
      });
```

src/prisma/migrations/20240327202800_add_keypairs/migration.sql
```diff
@@ -0,0 +1,8 @@
+-- CreateTable
+CREATE TABLE "keypairs" (
+    "hash" TEXT NOT NULL,
+    "publicKey" TEXT NOT NULL,
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
```
jakeloo, line 6: updatedAt?

src/prisma/schema.prisma
```diff
@@ -265,3 +265,12 @@ model ChainIndexers {
 
   @@map("chain_indexers")
 }
+
+model Keypairs {
+  hash      String   @id
+  publicKey String
+  algorithm String
+  createdAt DateTime @default(now())
```
jakeloo, line 273: no updatedAt?

src/server/routes/auth/keypair/import.ts
```diff
@@ -0,0 +1,83 @@
...
+      response: {
+        ...standardResponseSchema,
+        [StatusCodes.OK]: ReplySchema,
+      },
+    },
+    handler: async (req, res) => {
+      const { publicKey, algorithm } = req.body;
+
+      if (
+        !publicKey.startsWith("-----BEGIN PUBLIC KEY-----\n") ||
```
jakeloo, line 49: is this always consistent? 
is there a library or better way to do the validation based off the algorithm rather than hardcoding the string?

src/server/routes/auth/keypair/import.ts
```diff
@@ -0,0 +1,83 @@
...
+      tags: ["Keypair"],
+      operationId: "import",
+      body: BodySchema,
+      response: {
+        ...standardResponseSchema,
+        [StatusCodes.OK]: ReplySchema,
+      },
+    },
+    handler: async (req, res) => {
+      const { publicKey, algorithm } = req.body;
```
jakeloo, line 46: do we need to validate algorithm or we support all algorithm?

src/server/middleware/auth.ts
```diff
@@ -1,82 +1,47 @@
...
 import { getAuthWallet } from "../../utils/cache/authWallet";
 import { getConfig } from "../../utils/cache/getConfig";
 import { getWebhook } from "../../utils/cache/getWebhook";
+import { getKeypair } from "../../utils/cache/keypair";
 import { env } from "../../utils/env";
 import { logger } from "../../utils/logger";
 import { sendWebhookRequest } from "../../utils/webhook";
 import { Permission } from "../schemas/auth";
 
+const KEYPAIR_AUTH_MAX_DURATION_SECONDS = 15 * 60;
```
jakeloo, line 27: should be configurable per keypair probably?

src/server/middleware/auth.ts
```diff
@@ -176,46 +142,197 @@ export const onRequest = async ({
...
+    }
+
+    const keypair = await getKeypair({ publicKey: payload.iss });
+    if (!keypair || keypair.publicKey !== payload.iss) {
+      error =
+        "The provided public key is not configured for this Engine instance.";
+      throw error;
+    }
+
+    const { exp, iat } = jsonwebtoken.verify(jwt, keypair.publicKey, {
```
jakeloo, line 296: https://www.npmjs.com/package/jsonwebtoken#jwtverifytoken-secretorpublickey-options-callback

`issuer`, `clockTolerance`, etc. i think it checks expiration already?

jakeloo: since expiresAt should be encoded in the JWT?

src/server/routes/auth/keypair/import.ts
arcoraven: The Typebox input schema enforces one of:
 "RS256",
  "RS384",
  "RS512",
  "ES256",
  "ES384",
  "ES512",
  "PS256",
  "PS384",
  "PS512",

src/server/routes/auth/keypair/import.ts
arcoraven: Eh yeah I don't like this either. I tested each of the RS*/ES*/PS* and they all started with `-----BEGIN PUBLIC KEY-----`.

arcoraven: Actually this works quite well:
```
  function isWellFormedPublicKey(publicKeyString) {
    try {
      crypto.createPublicKey(publicKeyString);
      return true;
    } catch (e) {
      return false;
    }
  }
```

src/server/middleware/auth.ts
arcoraven: This was an explicit way to check that you don't provide an expiration that's too long, to prevent a footgun where someone sets a 5-year expiration (might as well use the permanent access tokens).

If it doesn't seem necessary I can remove this.

src/server/middleware/auth.ts
arcoraven: Yeah the original intent is no longer very valuable, will remove.

farhanW3 (approved):

PR #542, opened 2024-05-31
[WIP] Update: Key-Pair auth data check

farhanW3:
## PR-Codex overview
The focus of this PR is to enhance authentication middleware in the server by adding keypair authentication with body hashing validation.

### Detailed summary
- Added `createHash` import for hashing
- Updated hook from `onRequest` to `preValidation`
- Modified `handleKeypairAuth` to include body hashing validation
- Added `hashRequestBody` function to hash request body

### DIFF
--- a/src/server/middleware/auth.ts
+++ b/src/server/middleware/auth.ts
@@ -164,7 +164,7 @@ export const onRequest = async ({
       } else if (payload.iss === THIRDWEB_DASHBOARD_ISSUER) {
         return await handleDashboardAuth(jwt);
       } else {
-        return await handleKeypairAuth(jwt, payload.iss, req, payload.data);
+        return await handleKeypairAuth(jwt, req, payload.iss);
       }
     }
   }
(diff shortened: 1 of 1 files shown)


src/server/middleware/auth.ts
```diff
@@ -278,6 +282,11 @@ const handleKeypairAuth = async (
       throw error;
     }
 
+    if (data && req.method === "POST" && data !== hashRequestBody(req)) {
+      error = "The request body has been tampered with.";
+      throw error;
+    }
```
farhanW3, line 288: We can only do this check when the request is of POST type, otherwise the req.body will come as empty

src/server/middleware/auth.ts
```diff
@@ -278,6 +282,11 @@ const handleKeypairAuth = async (
       throw error;
     }
 
+    if (data && req.method === "POST" && data !== hashRequestBody(req)) {
+      error = "The request body has been tampered with.";
```
arcoraven, line 286: This error sounds scary when the dev may have just made a mistake. Let's make sure errors directly state the issue and optionally provide a way to fix it:

> The request body does not match the hash in the access token. See: https://portal.thirdweb.com/engine/features/keypair-authentication

farhanW3: Makes sense.

can we tinyurl this?

src/server/middleware/auth.ts
farhanW3: Added the `POST` check as `GET` spec doesn't support body as per the HTTP/1.1. Some web servers (Apache, Nginx & Traefik) drop the body param for GET request when forwarding the request. The spec doesn't forbid it, but the HTTP servers do drop it.

arcoraven (approved):

PR #546, opened 2024-06-14
Fix Error Parsing to Support Chain Overrides

pajserb0y:
## Changes

- Try relaying concurrent transactions signed by the same wallet while on the same non-commercial chain, the UNPREDICTABLE_GAS Error gets triggered, and when parsing Error data the getChainByChainIdAsync gets called but it doesn't have info about custom chain override it errors which results in infinite looping of all transactions trying to process and never do

- Here are the changes made.
- errors.ts added check for chainOverrides

## How this PR will be tested

- [ ] Open the dashboard and click X. Result: Try relaying concurrent transactions signed by the same wallet on custom chain override

## Output

---

## PR-Codex overview
The focus of this PR is to enhance error handling and configuration retrieval in the `errors.ts` file.

### Detailed summary
- Added `Chain` import
- Added `getConfig` import
- Updated error handling logic to use configuration overrides
- Improved error message readability

### DIFF
--- a/src/utils/errors.ts
+++ b/src/utils/errors.ts
@@ -1,11 +1,15 @@
 import { Transactions } from ".prisma/client";
-import { getChainByChainIdAsync } from "@thirdweb-dev/chains";
+import {
+  Chain,
+  getChainByChainIdAsync,
+} from "@thirdweb-dev/chains";
 import { ethers } from "ethers";
 import {
   createThirdwebClient,
   prepareTransaction,
   simulateTransaction,
 } from "thirdweb";
+import { getConfig } from "./cache/getConfig";
 import { env } from "./env";
 
 interface EthersError {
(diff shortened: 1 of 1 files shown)


arcoraven:
Hey there and thank you for the contribution! You're absolutely right that we aren't accounting for custom chain overrides in this error checker here.

Ideally we can abstract the chain resolver to a helper function instead since we'll likely want this "get chain from overrides or SDK" logic everywhere.

Is this actively blocking you right now? Are you able to delete the bad txs from the DB manually to unblock yourself?

pajserb0y:
Hey,

Not not acively after the fix I haven’t had any more problems m, and yes I
had to cancel manually the bad txns from DB, now everything that I am using
seems to be working fine,

Best,
Marko

PR #548, opened 2024-06-17
chore: support AA addresses in /send-transaction

arcoraven:
Support the `x-account-address` header in `/backend-wallet/:chain/send-transaction`

---

## PR-Codex overview
The focus of this PR is to update the wallet header schema to include an additional field `x-account-address` for sending transactions. 

### Detailed summary
- Updated `walletHeaderSchema` to `walletWithAAHeaderSchema`
- Added `x-account-address` field in the header for account address
- Modified transaction queuing based on the presence of `accountAddress` in the header

### DIFF
--- a/src/server/routes/backend-wallet/sendTransaction.ts
+++ b/src/server/routes/backend-wallet/sendTransaction.ts
@@ -8,7 +8,7 @@ import {
   transactionWritesResponseSchema,
 } from "../../schemas/sharedApiSchemas";
 import { txOverridesSchema } from "../../schemas/txOverrides";
-import { walletHeaderSchema } from "../../schemas/wallet";
+import { walletWithAAHeaderSchema } from "../../schemas/wallet";
 import { getChainIdFromChain } from "../../utils/chain";
 
 const ParamsSchema = Type.Object({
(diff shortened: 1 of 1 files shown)


src/server/routes/backend-wallet/sendTransaction.ts
```diff
@@ -71,19 +71,37 @@ export async function sendTransaction(fastify: FastifyInstance) {
...
-        idempotencyKey,
-        ...txOverrides,
-      });
+      let queueId: string;
+      if (accountAddress) {
+        const { id } = await queueTxRaw({
+          chainId: chainId.toString(),
+          signerAddress: fromAddress,
+          accountAddress,
+          target: toAddress,
```
joaquim-verges, line 84: arent these all optional? prob dont need the if / else right? just pass everything through even if undefined

arcoraven: There's some code like this:
```typescript
if (tx.accountAddress && tx.signerAddress) {
  // process as a userOp
}
```

In this case accountAddress is undefined so it would be okay, but I'm concerned about any stray code that tries to handle a userOp if signerAddress is set (and in this case we'd be setting signerAddress for EOA transactions too).

joaquim-verges (approved):

PR #569, opened 2024-07-19
Updates: Engine Error Handling

farhanW3:
Helps with better error message response when Zod error or SDK (etherjs) error codes

---

## PR-Codex overview
The focus of this PR is to update schemas with detailed descriptions and add error handling for Zod and Ethers errors.

### Detailed summary
- Updated NFT input and metadata schemas with detailed descriptions
- Added error handling for Zod and Ethers errors in the server middleware
- Defined ZodError and ETHERS_ERROR_CODES for error identification

### DIFF
--- a/src/server/middleware/error.ts
+++ b/src/server/middleware/error.ts
@@ -1,5 +1,6 @@
 import { FastifyInstance } from "fastify";
 import { ReasonPhrases, StatusCodes } from "http-status-codes";
+import { ZodError } from "zod";
 import { env } from "../../utils/env";
 import { logger } from "../../utils/logger";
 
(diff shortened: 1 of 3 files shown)


src/server/middleware/error.ts
```diff
@@ -20,6 +21,31 @@ export const createCustomError = (
   code,
 });
 
+const ETHERS_ERROR_CODES = [
```
arcoraven, line 24: If you have a source for this list, can you link it in a comment?

src/server/middleware/error.ts
```diff
@@ -31,42 +57,99 @@ export const createCustomDateTimestampError = (key: string): CustomError => {
...
       });
-    }
-  });
+
+      // Ethers Error Codes
+      if (isEthersError(error)) {
+        return reply.status(StatusCodes.BAD_REQUEST).send({
+          error: {
+            code: "BAD_REQUEST",
+            message: (error as any).code,
```
arcoraven, line 90: This is type unsafe. Can we avoid `as` here? Maybe something like:
```typescript
message: "code" in error ? error.code : error.message
```

arcoraven (approved):
Hmm in general I think the error middleware has too much logic in it, but this solves an immediate need. In the spirit of keeping code more readable, I feel like it could use a rework to identify the error type, extract the right info from it, and DRY some repeated logic.

PR #595, opened 2024-08-01
Fk/extend ph worker

farhanW3:
## PR-Codex overview
The focus of this PR is to enhance the nonce management in the transaction sending process. 

### Detailed summary
- Added `rebaseNonce` function to resync the nonce based on transaction count.
- Updated `sendTransactionWorker.ts` to handle nonce resync on `NonceAlreadyUsedError`.
- Exported `lastUsedNonceKey` function in `walletNonce.ts`.


src/worker/tasks/sendTransactionWorker.ts
```diff
@@ -167,6 +171,11 @@ const _sendTransaction = async (
     transactionHash = sendTransactionResult.transactionHash;
     job.log(`Sent transaction: ${transactionHash}`);
   } catch (error: unknown) {
+    // If NonceAlreadyUsedError, rebase the nonce and retry.
+    if (isNonceAlreadyUsedError(error)) {
+      const resyncNonce = await syncNonce(chainId, from);
+      job.log(`Resynced nonce to ${resyncNonce}.`);
+    }
```
arcoraven, line 178: Change the line below this to `} else {`

src/db/wallets/walletNonce.ts
```diff
@@ -122,3 +122,36 @@ export const deleteAllNonces = async () => {
     await redis.del(keys);
   }
 };
+
+/**
+ * Resync the nonce i.e., max of transactionCount +1 and lastUsedNonce.
+ * @param chainId
+ * @param walletAddress
+ */
+export const rebaseNonce = async (chainId: number, walletAddress: Address) => {
```
arcoraven, line 131: This isn't used. Did you mean to call it somewhere?

arcoraven, line 131: Since there's no difference if this func is called more than once, we can put a lock around this and exit if the lock is taken.

Pseudocode:
```
acquire a lock for 10 seconds
if the lock is acquired
  query rpc for transactionCount
  update redis
  release lock
```

arcoraven: Actually with a lock, do we even need a lua script?

d4mr (approved):

PR #597, opened 2024-08-03
Updates: v2 Nonce

farhanW3:
Added `nonceResyncQueue` & `nonceResyncWorker`.

New flows:
- on send completion, we insert the nonce into an set `sentnonce:<wallet>:chain`
- on mine completion, 
  - if tx mined, we remove the nonce from the set `sentnonce:<wallet>:chain`
  - if nonce is not mined & retry > maxRetries & not in recycle-nonce Q:
    - update the queueId with `errored` state
    - we add the nonce to recycle Q
      - either it will be cancelled or consumed
- new `nonceResync` cron :
  - checks sentnonce* keys to get the wallets in use &  gets onchain nonce & dbNonce
  - if onchain_nonce - 1 < dbCount :
    - if missing nonce(s) is not in recycle-nonce Q
     - we put the nonce in recycle-nonce Q // this step catches the nonces

with this we catch all missing nonce

---

## PR-Codex overview
This PR focuses on implementing a `NonceResyncWorker` to periodically resync nonces for wallets, improving nonce management in the system.

### Detailed summary
- Added `NonceResyncWorker` to periodically check and recycle missing nonces for wallets.
- Introduced `sentNoncesKey`, `addSentNonce`, `removeSentNonce`, and `isSentNonce` functions for managing sent nonces.
- Modified several files to incorporate the new nonce management functionality.

> The following files were skipped due to too many changes: `src/worker/tasks/nonceResyncWorker.ts`

### DIFF
--- a/src/db/wallets/walletNonce.ts
+++ b/src/db/wallets/walletNonce.ts
@@ -1,5 +1,6 @@
 import { Address, eth_getTransactionCount, getRpcClient } from "thirdweb";
 import { getChain } from "../../utils/chain";
+import { logger } from "../../utils/logger";
 import { normalizeAddress } from "../../utils/primitiveTypes";
 import { redis } from "../../utils/redis/redis";
 import { thirdwebClient } from "../../utils/sdk";
@@ -58,6 +59,14 @@ export const recycleNonce = async (
   walletAddress: Address,
   nonce: number,
 ) => {
+  if (isNaN(nonce)) {
+    logger({
+      level: "warn",
+      message: `[recycleNonce] Invalid nonce: ${nonce}`,
+      service: "worker",
+    });
+    return;
+  }
   const key = recycledNoncesKey(chainId, walletAddress);
   await redis.rpush(key, nonce);
 };
--- a/src/db/wallets/walletNonce.ts
+++ b/src/db/wallets/walletNonce.ts
@@ -59,7 +59,8 @@ export const removeSentNonce = async (
   nonce: number,
 ) => {
   const key = sentNoncesKey(chainId, walletAddress);
-  await redis.srem(key, nonce.toString());
+  const removed = await redis.srem(key, nonce.toString());
+  return removed === 1;
 };
 
 /**
--- a/src/db/wallets/walletNonce.ts
+++ b/src/db/wallets/walletNonce.ts
@@ -15,7 +15,7 @@ export const lastUsedNonceKey = (chainId: number, walletAddress: Address) =>
  * The "recycled nonces" list stores unsorted nonces to be reused or cancelled.
  * Example: [ "25", "23", "24" ]
  */
-const recycledNoncesKey = (chainId: number, walletAddress: Address) =>
+export const recycledNoncesKey = (chainId: number, walletAddress: Address) =>
   `nonce-recycled:${chainId}:${normalizeAddress(walletAddress)}`;
 
 /**
@@ -138,12 +138,17 @@ export const rebaseNonce = async (chainId: number, walletAddress: Address) => {
   const transactionCount = await eth_getTransactionCount(rpcRequest, {
     address: walletAddress,
   });
+  console.log("::Debug Log:: transactionCount: ", transactionCount);
+  console.log(
+    "::Debug Log:: lastUsedNonceKey Value: ",
+    await redis.get(lastUsedNonceKey(chainId, normalizeAddress(walletAddress))),
+  );
 
-  // Lua script to set nonce as max(transactionCount, redis.get(lastUsedNonceKey))
+  // Lua script to set nonce as max
   const script = `
     local transactionCount = tonumber(ARGV[1])
     local lastUsedNonce = tonumber(redis.call('get', KEYS[1]))
-    local nextNonce = math.max(transactionCount, lastUsedNonce)
+    local nextNonce = math.max(transactionCount-1, lastUsedNonce)
     redis.call('set', KEYS[1], nextNonce)
     return nextNonce
   `;
--- a/src/db/wallets/walletNonce.ts
+++ b/src/db/wallets/walletNonce.ts
@@ -138,11 +138,6 @@ export const rebaseNonce = async (chainId: number, walletAddress: Address) => {
   const transactionCount = await eth_getTransactionCount(rpcRequest, {
     address: walletAddress,
   });
-  console.log("::Debug Log:: transactionCount: ", transactionCount);
-  console.log(
-    "::Debug Log:: lastUsedNonceKey Value: ",
-    await redis.get(lastUsedNonceKey(chainId, normalizeAddress(walletAddress))),
-  );
 
   // Lua script to set nonce as max
   const script = `
--- a/src/db/wallets/walletNonce.ts
+++ b/src/db/wallets/walletNonce.ts
@@ -63,6 +63,18 @@ export const removeSentNonce = async (
 };
 
 /**
+ * Check if a nonce is in the sent nonces set.
+ */
+export const isSentNonce = async (
+  chainId: number,
+  walletAddress: Address,
+  nonce: number,
+) => {
+  const key = sentNoncesKey(chainId, walletAddress);
+  return !!(await redis.sismember(key, nonce.toString()));
+};
+
+/**
  * Acquire an unused nonce.
  * This should be used to send an EOA transaction with this nonce.
  * @param chainId
--- a/src/db/wallets/walletNonce.ts
+++ b/src/db/wallets/walletNonce.ts
@@ -13,11 +13,54 @@ export const lastUsedNonceKey = (chainId: number, walletAddress: Address) =>
   `nonce:${chainId}:${normalizeAddress(walletAddress)}`;
 
 /**
- * The "recycled nonces" list stores unsorted nonces to be reused or cancelled.
+ * The "recycled nonces" set stores unsorted nonces to be reused or cancelled.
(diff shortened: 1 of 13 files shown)


src/worker/tasks/mineTransactionWorker.ts
```diff
@@ -128,6 +128,19 @@ const _mineTransaction = async (
     if (result.status === "fulfilled") {
       const receipt = result.value;
       job.log(`Found receipt on block ${receipt.blockNumber}.`);
+
+      // removed nonce from sentnonce set
+      const removed = await redis.srem(
+        `sentnonce:${receipt.from.toLowerCase()}:${chainId}`,
```
d4mr, line 134: we should create a helper like in the case of `recycledNoncesKey()` to fetch the key everywhere

d4mr:
added a comment, also it would be good to convert `recycledNonces` to a set. Because there is another source now where we push to `recycledNonces`, and we do this non-atomically, having it be a set would be useful.

d4mr (changes requested):

src/worker/tasks/mineTransactionWorker.ts
arcoraven: Can we use this pattern `nonce-sent:1993:0x...`

src/db/wallets/walletNonce.ts
```diff
@@ -58,8 +102,16 @@ export const recycleNonce = async (
   walletAddress: Address,
   nonce: number,
 ) => {
+  if (isNaN(nonce)) {
```
arcoraven, line 105: I've sometimes seen NaN nonces in the recycled nonces list too. Were you able to find out why? Is this log getting printed?

src/worker/tasks/mineTransactionWorker.ts
```diff
@@ -160,6 +173,32 @@ const _mineTransaction = async (
...
+      sentTransaction.chainId,
+      sentTransaction.from,
+      sentTransaction.nonce,
+    );
+
+    await recycleNonce(
+      sentTransaction.chainId,
+      sentTransaction.from,
+      sentTransaction.nonce,
+    );
```
arcoraven, line 201: This shouldn't be here. All this logic already exists on L233 of this file in the "failed" listener after the last attempt. Effectively when a job tries the max num of attempts, it is considered "timed out" and:
- errors the tx
- sends the error webhook/usage event
- recycles the nonce
You just need to add the `removeSentNonce` step there.

test/e2e/tests/setup.ts
```diff
@@ -28,6 +28,7 @@ export const setup = async (): Promise<SetupResult> => {
 
   const publicClient = setupPublicClient();
   const engine = setupEngine();
+  // const backendWallet = "0x7c243d1475da6fe0eeb8d1c4ee4510af35cf7cd5";
```
arcoraven, line 31: not sure if these comments were intended to stay here

src/worker/tasks/nonceResyncWorker.ts
```diff
@@ -0,0 +1,103 @@
...
+export const sleep = async (ms: number) =>
+  new Promise((resolve) => setTimeout(resolve, ms, null));
+
+/**
+ * Resyncs nonces for all wallets.
+ * This worker should be run periodically to ensure that nonces are not skipped.
+ * It checks the onchain nonce for each wallet and recycles any missing nonces.
+ */
+const handler: Processor<any, void, string> = async (job: Job<string>) => {
+  const sentNoncesKeys = await redis.keys("*nonce-sent*");
```
arcoraven, line 42: Use `nonce-sent:*` to be stricter

src/worker/tasks/nonceResyncWorker.ts
```diff
@@ -0,0 +1,103 @@
...
+    const transactionCount = await eth_getTransactionCount(rpcRequest, {
+      address: walletAddress,
+    });
+
+    // Get DB Nonce
+    const dbNonceCount = Number(
+      await redis.get(lastUsedNonceKey(chainId, walletAddress)),
+    );
+    job.log(
+      `[wallet ${walletAddress}] onchain Nonce: ${transactionCount} and DBNonce: ${dbNonceCount}`,
```
arcoraven, line 64: Nits:
- the word `wallet` isn't useful since we know from context, and this is internal only
- lowercase `nonce`, it's not a proper noun or product name
- `DBNonce` -> `db nonce` for consistency: `"onchain nonce 5 and db nonce 6"`

d4mr (approved):

src/worker/tasks/nonceResyncWorker.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    // The next unused nonce = transactionCount.
+    const transactionCount = await eth_getTransactionCount(rpcRequest, {
+      address: walletAddress,
+    });
+
+    // Get DB Nonce
+    const dbNonceCount = Number(
+      await redis.get(lastUsedNonceKey(chainId, walletAddress)),
+    );
```
arcoraven, line 62: There's an existing helper func for this: `inspectNonce(...)`

Also these var names seem confusing and this is a very easy thing to mix up:
- The transactionCount is the _number of transactions sent from this wallet_. 5 means this wallet has sent 5 transactions, the most recent being nonce 4.
- The last used nonce Redis key stores the _last used nonce value for a wallet_. 5 means this wallet already sent nonce 5 and has sent 6 transactions.

We shouldn't compare the two values since the former is the "next unused nonce" and the latter is the "last used nonce".

To make it more clear, let's try to consistently name them and document them. Let's pick last used nonce.
- `lastUsedNonceOnchain = transactionCount - 1`
- `lastUsedNonceDB = inspectNonce(...)`

src/worker/tasks/nonceResyncWorker.ts
```diff
@@ -0,0 +1,103 @@
...
+    job.log(
+      `[wallet ${walletAddress}] onchain Nonce: ${transactionCount} and DBNonce: ${dbNonceCount}`,
+    );
+    logger({
+      level: "debug",
+      message: `[nonceResyncWorker] onchain Nonce: ${transactionCount} and DBNonce: ${dbNonceCount}`,
+      service: "worker",
+    });
+
+    if (transactionCount >= dbNonceCount + 1) {
```
arcoraven, line 72: Isn't this a bad condition that needs correcting?
The tx count is ahead of the redis nonce + 1. This means an increment will return an already used nonce.

Example:
- transactionCount = 5
- dbNonceCount = 2
- 5 is >= 2 + 1
- The next increments will be 3, 4, and 5 which are all already used nonces.

Do you plan to rely on the incr retries to fix this case? It seems like we could rebase the nonce here.

src/worker/tasks/nonceResyncWorker.ts
```diff
@@ -0,0 +1,103 @@
...
+      });
+      return;
+    }
+
+    // Check if nonce exists in nonce-sent set
+    // If not, recycle nonce
+    for (let _nonce = transactionCount; _nonce < dbNonceCount; _nonce++) {
+      if (isNaN(_nonce)) {
+        continue;
+      }
```
arcoraven, line 87: What is this check for? nonce should never be NaN right?

src/worker/tasks/nonceResyncWorker.ts
```diff
@@ -0,0 +1,103 @@
...
+      const exists = await redis.sismember(sentNonceKey, _nonce.toString());
```
arcoraven, line 88: This could be a helper func

src/db/wallets/walletNonce.ts
farhanW3: I saw it too

src/worker/tasks/mineTransactionWorker.ts
d4mr: thanks, good catch. wonder how a test could have been setup to catch this, since it's very criticial

src/worker/tasks/nonceResyncWorker.ts
d4mr: this cron worker only handles the problem of chain nonce being stuck behind db nonce.
if the chain nonce is ahead of the db nonce, we should get a nonce too low, and we have a resolution in place within the sendTransaction worker to fix that

src/worker/tasks/nonceResyncWorker.ts
d4mr: true, but we depend on RPC for the onchain transaction count. While it should never return a string, better safe than sorry?

src/worker/tasks/sendTransactionWorker.ts
```diff
@@ -171,8 +172,9 @@ const _sendTransaction = async (
     transactionHash = sendTransactionResult.transactionHash;
     job.log(`Sent transaction: ${transactionHash}`);
   } catch (error: unknown) {
-    // If NonceAlreadyUsedError, rebase the nonce and retry.
-    if (isNonceAlreadyUsedError(error)) {
+    // If NonceAlreadyUsedError, which can also manifest as a ReplacementGasFeeTooLowError,
+    // recycle the nonce and retry the transaction.
+    if (isNonceAlreadyUsedError(error) || isReplacementGasFeeTooLow(error)) {
```
farhanW3, line 177: We don't need `isReplacementGasFeeTooLow` here.

We can avoid this error totally, if we use the `max(onchain-gas, previous-tx-gas-values)` for gas values of the new transactions.

PR #615, opened 2024-08-22
Nonce and Queue Stats Log Cron

d4mr:
## PR-Codex overview
This PR focuses on adding a `NonceHealthCheckQueue` class and a worker to monitor nonce health for wallets.

### Detailed summary
- Added `NonceHealthCheckQueue` class to monitor nonce health
- Implemented a worker to check nonce status for wallets
- Added routes to get nonce status details for wallets

> The following files were skipped due to too many changes: `src/server/routes/admin/nonces.ts`


src/server/routes/admin/nonces.ts
```diff
@@ -0,0 +1,169 @@
...
+        description: "Last mined nonce",
+        examples: ["0"],
+      }),
+      lastUsedNonce: Type.String({
+        description: "Last incremented nonce sent to the blockchain",
+        examples: ["0"],
+      }),
+      sentNonces: Type.Array(
+        Type.String({
+          examples: ["0"],
```
arcoraven, line 35: I think you want `examples` on the `Array` and not the `String`?

src/server/routes/admin/nonces.ts
```diff
@@ -0,0 +1,169 @@
...
+        }),
+        {
+          description:
+            "Nonces that were successfully sent to the blockchain but not mined yet.",
```
arcoraven, line 39: I think we can say "sent to RPC" since "sent to the blockchain" might suggest it's onchain.

src/server/routes/admin/nonces.ts
```diff
@@ -0,0 +1,169 @@
...
+          examples: ["0"],
+        }),
+        {
+          description:
+            "Nonces that were acquired but failed to be sent to the blockchain, waiting to be recycled or cancelled.",
+        },
+      ),
+      chainId: Type.Number({
+        description: "Chain ID",
+        examples: [80001],
```
arcoraven, line 53: Use 80002 as an example, 80001 is Mumbai -> deprecated

arcoraven: Also nit: put this filed next to wallet address because the `chainId + address` tuple defines a wallet.

src/server/routes/admin/nonces.ts
```diff
@@ -0,0 +1,169 @@
...
+import { walletWithAddressParamSchema } from "../../schemas/wallet";
+
+export const responseBodySchema = Type.Object({
+  result: Type.Array(
+    Type.Object({
+      walletAddress: Type.String({
+        description: "Backend Wallet Address",
+        examples: ["0xcedf3b4d8f7f1f7e0f7f0f7f0f7f0f7f0f7f0f7f"],
+      }),
+      onchainNonce: Type.String({
```
arcoraven, line 25: IMO use numbers as nonces. They're already numbers throughout the code, they shouldn't be larger than INT_MAX, and they're more sort-able.

src/server/routes/admin/nonces.ts
```diff
@@ -0,0 +1,169 @@
...
+    Querystring: Static<typeof walletWithAddressParamSchema>;
+    Reply: Static<typeof responseBodySchema>;
+  }>({
+    method: "GET",
+    url: "/admin/nonces",
+    schema: {
+      summary: "Get transaction details",
+      description: "Get raw logs and details for a transaction by queueId.",
+      tags: ["Admin"],
+      operationId: "transactionDetails",
```
arcoraven, line 82: Update these

src/server/routes/admin/nonces.ts
```diff
@@ -0,0 +1,169 @@
...
+  const result = await Promise.all(
+    lastUsedNonceKeys.map(async (key) => {
+      const { chainId, walletAddress } = splitLastUsedNonceKey(key);
+
+      const [lastUsedNonce, sentNonces, recycledNonces, onchainNonce] =
+        await Promise.all([
+          redis.get(lastUsedNonceKey(chainId, walletAddress)),
+          redis.smembers(sentNoncesKey(chainId, walletAddress)),
+          redis.smembers(recycledNoncesKey(chainId, walletAddress)),
+          getOnchainNonce(chainId, walletAddress),
```
arcoraven, line 119: This is fine. Another thing you could do is redis pipelines which batch send reqs to Redis in one network call sequentially and return an array. This makes 4 network calls.

src/server/routes/admin/nonces.ts
```diff
@@ -0,0 +1,169 @@
...
+        ]);
+
+      return {
+        walletAddress,
+        onchainNonce: onchainNonce.toString(),
+        lastUsedNonce: lastUsedNonce ?? "0",
+        sentNonces,
+        recycledNonces,
```
arcoraven, line 127: Would be nice to sort these, hence using numbers

src/worker/queues/logQueue.ts
```diff
@@ -0,0 +1,14 @@
+import { Queue } from "bullmq";
+import { redis } from "../../utils/redis/redis";
+import { defaultJobOptions } from "./queues";
+
+export class LogQueue {
```
arcoraven, line 5: Can we use a clearer name? I have no idea what this does based on the name? Ideally something like `<verb><noun>Queue`

src/server/routes/admin/nonces.ts
d4mr: will have to parse all of them because redis returns strings. I want to be consistent between redis nonce and onchain nonce so either all strings or all numbers. I am fine with doing all numbers, its just needing to parse the returned set.

src/server/routes/admin/nonces.ts
d4mr: makes sense. I would think offloading the compute to redis would be more efficient (?) so when we move to sorted sets we will receive them sorted anyway

src/server/routes/admin/nonces.ts
arcoraven: I think we should parse them then. I predict we'll want them sortable on the frontend. And visually (like when I curl the endpoint) I'd want to see them sorted too. Rather do that on the backend if it's going to be a common need.

src/server/routes/admin/nonces.ts
arcoraven: IMO don't worry about efficiency. string -> number is free. Worry about code cleanliness over performance.

Especially if we're going to sort in the dashboard. I'd rather our backend cast the type + sort than the client's browser.

src/server/routes/admin/nonces.ts
d4mr: agreed! more wallet/network combos add up to a lot more network calls. I have flattened it all out to a single pipeline

src/worker/queues/logQueue.ts
d4mr: renamed it to `nonceHealthCheckQueue`, open to suggestions

d4mr:
closing in favour of #628

PR #618, opened 2024-08-23
Updates: Claim-to using v5 SDK

farhanW3:
## PR-Codex overview
The focus of this PR is to update the contract interaction functions to use the new v5 SDK and improve transaction handling in ERC20, ERC721, and ERC1155 extensions.

### Detailed summary
- Updated contract interaction functions to use `getContractV5` with new v5 SDK
- Improved transaction handling with `insertTransaction` and `claimTo` functions
- Added `Address` and `Hex` imports from `thirdweb` for type safety

### DIFF
--- a/src/utils/cache/getContractv5.ts
+++ b/src/utils/cache/getContractv5.ts
@@ -1,0 +1,34 @@
+import { ContractOptions, defineChain, getContract } from "thirdweb";
+import { thirdwebClient } from "../../utils/sdk";
+
+interface GetContractParams {
+  chainId: number;
+  contractAddress: string;
+}
+
+// export const definedChainCache = new Map<number, Chain>(); // This is a cache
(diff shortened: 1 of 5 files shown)


src/utils/cache/getContractv5.ts
```diff
@@ -0,0 +1,34 @@
...
+  // get a contract
+  return getContract({
+    // the client you have created via `createThirdwebClient()`
+    client: thirdwebClient,
+    // the contract's address
+    address: contractAddress,
+    // the chain the contract is deployed on
+    chain: definedChain,
+  });
+};
```
arcoraven, line 34: We shouldn't need to cache v5 contracts. I believe I did a quick test before: querying the same contract multiple time is already cached (as long as you use the same `client`). We can remove all the caching.

farhanW3: Got it. Will update

PR #626, opened 2024-08-26
feat: migrate aws kms flow to v5

d4mr:
## PR-Codex overview
This PR focuses on enhancing wallet management by introducing AWS KMS and GCP KMS wallet functionalities, improving configuration handling, and updating schemas and tests to support the new features.

### Detailed summary
- Removed unused files: `getAwsKmsWallet.ts`, `getGcpKmsWallet.ts`.
- Introduced AWS KMS and GCP KMS wallet management functions.
- Updated environment variables for AWS and GCP configurations.
- Enhanced error handling for wallet creation and fetching parameters.
- Modified `createWalletDetails` to handle new wallet types.
- Added new utility functions for splitting and constructing KMS resource paths.
- Updated database schema to accommodate new wallet parameters.
- Improved test coverage for AWS KMS and GCP KMS functionalities.

> The following files were skipped due to too many changes: `src/prisma/schema.prisma`, `src/db/configuration/getConfiguration.ts`, `src/utils/account.ts`, `src/server/routes/backend-wallet/import.ts`, `yarn.lock`


test/e2e/tests/setup.ts
```diff
@@ -28,7 +23,8 @@ export const setup = async (): Promise<SetupResult> => {
 
   const publicClient = setupPublicClient();
   const engine = setupEngine();
-  const backendWallet = await getEngineBackendWallet(engine);
+  // const backendWallet = await getEngineBackendWallet(engine);
+  const backendWallet = "0x50495dacda9553e50dad15cd3a3016743f71b4c9";
```
arcoraven, line 27: Did you mean for this change?

src/server/utils/wallets/getAwsKmsAccount.ts
```diff
@@ -0,0 +1,120 @@
+import { type KMSClientConfig } from "@aws-sdk/client-kms";
+import { KmsSigner } from "aws-kms-signer";
+import type { Hex, ThirdwebClient } from "thirdweb";
+import { keccak256, type Address as TwAddress } from "thirdweb";
```
arcoraven, line 4: Nit: Why override the name here?

src/server/utils/wallets/getAwsKmsAccount.ts
```diff
@@ -0,0 +1,120 @@
...
+};
+
+export async function getAwsKmsAccount(
+  options: AwsKmsAccountOptions,
+): Promise<Account> {
+  const { keyId, config, client } = options;
+  const signer = new KmsSigner(keyId, config);
+
+  // Populate address immediately
+  const addressUnprefixed = await signer.getAddress();
```
arcoraven, line 39: This seems weird the underlying lib returns the address unprefixed. Is that documented somewhere?

src/server/utils/wallets/getAwsKmsAccount.ts
```diff
@@ -0,0 +1,120 @@
...
+  const address = `0x${addressUnprefixed}` as TwAddress;
+
+  return {
+    address,
+
+    sendTransaction: async (
```
arcoraven, line 45: Nit: these are meaty functions. Can we break them out into functions?

src/server/utils/wallets/getAwsKmsAccount.ts
d4mr: don't think so, documentation is scarce, I mostly went by interface and source code

test/e2e/tests/setup.ts
d4mr: nope, reverted. This one keeps falling through lol, high time to make it configurable properly.

src/server/utils/wallets/getAwsKmsAccount.ts
d4mr: remnant of a different file where it caused a conflict, but now the conflicting import is gone. Fixing.

src/server/utils/wallets/getAwsKmsAccount.ts
d4mr: they rely on in-scope values, so separating thme outside of the `getAwsKmsAccount` scope with extra params seems unnecessary. Can move the definition within the body of `getAwsKmsAccount` though, so the returned object is simpler to read

src/server/utils/wallets/getAwsKmsAccount.ts
```diff
@@ -0,0 +1,120 @@
...
+        Buffer.from(typedDataHash.slice(2), "hex"),
+      );
+      return `0x${signature.toString()}`;
+    },
+
+    signTransaction: async (tx: TransactionSerializable): Promise<Hex> => {
+      const serializedTx = serializeTransaction({ transaction: tx });
+      const txHash = keccak256(serializedTx);
+      const signature = await signer.sign(Buffer.from(txHash.slice(2), "hex"));
+      return `0x${signature.toString()}`;
```
joaquim-verges, line 117: i was expecting to see this re-used in sendTransaction, but see that flow is a bit different between the 2?

I think they should be the same, signTransaction should return the 'broadcastable' transaction, so the output of `serializeTransaction`.

if you assign the whole account object to a const, like `const account = { address, sendTransaction, ... }`

then you can reference it from inside the object, so you can do `account.signTransaction()` inside `sendTransaction`

d4mr: fixed this!

src/server/utils/wallets/getAwsKmsAccount.ts
d4mr: I moved it inside the body, although after the change I think having it in the return object was more readable to me.
Let me know if you'd like me to revert it or something else altogether

src/server/utils/wallets/getAwsKmsAccount.ts
```diff
@@ -0,0 +1,122 @@
...
+    tx: SendTransactionOption,
+  ): Promise<SendTransactionResult> {
+    const rpcRequest = getRpcClient({
+      client: client,
+      chain: await getChain(tx.chainId),
+    });
+
+    const signedTx = await signTransaction(tx);
+
+    const transactionHash = await eth_sendRawTransaction(rpcRequest, signedTx);
```
joaquim-verges, line 111: clean 👌

joaquim-verges (approved):
lgtm

joaquim-verges:
@d4mr lets get this in ser

src/db/wallets/createWalletDetails.ts
```diff
@@ -8,13 +9,23 @@ interface CreateWalletDetailsParams {
   address: string;
   type: WalletType;
   label?: string;
-  awsKmsKeyId?: string;
+
+  // AWS KMS
+  awsKmsKeyId?: string; // depcrecated and unused, todo: remove with next breaking change
```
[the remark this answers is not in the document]
arcoraven:
```typescript
import { STSClient, GetCallerIdentityCommand } from "@aws-sdk/client-sts";

const stsClient = new STSClient({});
const data = await stsClient.send(new GetCallerIdentityCommand({}));
const accountId = data.Account;
```

Not sure if it's worth doing this now since this PR is quite large already, unless you think it simplifies code complexity.

src/prisma/schema.prisma
```diff
@@ -76,20 +76,24 @@ model Tokens {
...
   // Local
-  encryptedJson      String? @map("encryptedJson")
+  encryptedJson                      String? @map("encryptedJson")
   // KMS
-  awsKmsKeyId        String? @map("awsKmsKeyId")
-  awsKmsArn          String? @map("awsKmsArn")
+  awsKmsKeyId                        String? @map("awsKmsKeyId") // deprecated and unused, todo: remove with next breaking change. Use awsKmsArn
+  awsKmsArn                          String? @map("awsKmsArn")
+  awsKmsSecretAccessKey              String? @map("awsKmsSecretAccessKey") // if not available, default to: Configuration.awsSecretAccessKey
+  awsKmsAccessKeyId                  String? @map("awsKmsAccessKeyId") // if not available, default to: Configuration.awsAccessKeyId
```
arcoraven, line 88: I'm not sure if the ROI on letting users set different configuration per wallet is worth the code/DX complexity. While there are use cases this supports, I don't think anyone has saw this as a problem, and the workaround is to make sure your creds have permissions on all regions.

src/prisma/schema.prisma
```diff
@@ -76,20 +76,24 @@ model Tokens {
...
   // GCP
-  gcpKmsKeyRingId    String? @map("gcpKmsKeyRingId") @db.VarChar(50)
-  gcpKmsKeyId        String? @map("gcpKmsKeyId") @db.VarChar(50)
-  gcpKmsKeyVersionId String? @map("gcpKmsKeyVersionId") @db.VarChar(20)
-  gcpKmsLocationId   String? @map("gcpKmsLocationId") @db.VarChar(20)
-  gcpKmsResourcePath String? @map("gcpKmsResourcePath") @db.Text
+  gcpKmsKeyRingId                    String? @map("gcpKmsKeyRingId") @db.VarChar(50) // deprecated and unused. Use gcpKmsResourcePath instead, todo: remove with next breaking change
```
arcoraven, line 90: Tip: You can use triple slash:
`/// @deprecated This is no longer tracked in DB.`

Prisma will generate it in typedocs for our editor:
image

src/server/routes/backend-wallet/import.ts
```diff
@@ -1,29 +1,48 @@
...
+        awsAccessKeyId: Type.String({ description: "AWS Access Key ID" }),
+        awsSecretAccessKey: Type.String({
+          description: "AWS Secret Access Key",
+        }),
+      },
+      {
+        description:
+          "Optional AWS credentials to use for importing the wallet, if not provided, the default AWS credentials will be used (if available).",
+      },
+    ),
```
arcoraven, line 37: As mentioned above, I'd rather only prompt the user for _metadata_ and not _credentials_

src/server/routes/backend-wallet/import.ts
```diff
@@ -1,29 +1,48 @@
...
 import { importGcpKmsWallet } from "../../utils/wallets/importGcpKmsWallet";
 import { importLocalWallet } from "../../utils/wallets/importLocalWallet";
 
 const RequestBodySchema = Type.Union([
   Type.Object({
-    awsKmsKeyId: Type.String({
-      description: "AWS KMS key ID",
-      examples: ["12345678-1234-1234-1234-123456789012"],
-    }),
+    label: Type.Optional(
```
arcoraven, line 15: Since they all have label, it might be simpler to move the label to the top level
```typescript
schema = {
  label?: string
} & (
  | { <local wallet config> }
  | { <aws-kms wallet config> }
  | { <gcp-kms wallet config> }
)
```

That might look like this:
```typescript
const schema = Type.Intersect([
  Type.Object({
    label: Type.Optional(Type.String()),
  }),
  Type.Union([
    Type.Object({
      privateKey: Type.String(),
      // ...
    }),
    Type.Object({
      awsArn: Type.String(),
      // ...
    }),
    Type.Object({
      gcpKmsKeyRingId: Type.String(),
      // ...
    })
  ])
])
```

src/server/routes/configuration/wallets/get.ts
```diff
@@ -43,10 +37,23 @@ export async function getWalletsConfiguration(fastify: FastifyInstance) {
         [StatusCodes.OK]: responseBodySchema,
       },
     },
-    handler: async (req, res) => {
+    handler: async (_req, res) => {
```
arcoraven, line 40: Any reason for this change? Looks unneeded

src/server/routes/configuration/wallets/update.ts
```diff
@@ -68,76 +63,66 @@ export async function updateWalletsConfiguration(fastify: FastifyInstance) {
...
+          awsSecretAccessKey: req.body.awsSecretAccessKey,
+          awsRegion: req.body.awsRegion,
+        });
+      }
 
-          await updateConfiguration({
-            awsAccessKeyId: req.body.awsAccessKeyId,
-            awsSecretAccessKey: req.body.awsSecretAccessKey,
-            awsRegion: req.body.awsRegion,
-            gcpApplicationProjectId: null,
```
arcoraven, line 101: Ok good, this was the main part that would delete other wallet creds when adding new configs.

package.json
```diff
@@ -28,6 +28,8 @@
   "dependencies": {
     "@aws-sdk/client-kms": "^3.398.0",
     "@bull-board/fastify": "^5.21.1",
+    "@cloud-cryptographic-wallet/cloud-kms-signer": "^0.1.2",
```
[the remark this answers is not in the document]
d4mr: The package comes from the monorepo you showed, AWS KMS signer was also in the same monorepo.

Although some packages there are built in ethers both the AWS and GCP signer packages are solely built on top of their respective SDKs, and have no ethers dependencies

src/prisma/schema.prisma
d4mr: Nice, very handy, thanks!

src/server/routes/configuration/wallets/get.ts
d4mr: biome complains if any variables are unused, unless they are underscore prefixed.

I'm trying to make every file I touched lint-error free

src/prisma/schema.prisma
jakeloo: It makes sense to have it in this layer of abstraction -- it's a good practice to provision granular scope access key id and secret, so in this case, we can support per kms arn wallet  = 1 aws key. instead of 1 aws key = engine with a wider scope

jakeloo: imo, lower-level code abstraction should provide us the flexibility to support wider use case. DX problems can be addresses through a higher level abstraction that glue lower-level interfaces together. This way we can get a good dx + being flexible

arcoraven (approved):

src/server/routes/backend-wallet/import.ts
```diff
@@ -1,39 +1,69 @@
...
-      examples: ["12345678-1234-1234-1234-123456789012"],
+  Type.Union([
+    Type.Object({
+      awsKmsArn: Type.String({
+        description: "AWS KMS key ARN",
+        examples: [
+          "arn:aws:kms:us-east-1:123456789012:key/12345678-1234-1234-1234-123456789012",
+        ],
+      }),
+      credentials: Type.Object(
```
arcoraven, line 29: I think there should be a `Type.Optional` here.

d4mr: thanks, good catch

src/server/utils/wallets/createAwsKmsWallet.ts
```diff
@@ -1,25 +1,42 @@
...
-    throw new Error(`Server was not configured for AWS KMS wallet creation`);
+  let kmsWalletParams: AwsKmsWalletParams;
+  try {
+    kmsWalletParams = await fetchAwsKmsWalletParams(overrides);
+  } catch (e) {
+    if (e instanceof FetchAwsKmsWalletParamsError) {
+      throw new CreateAwsKmsWalletError(e.message);
+    }
+    throw e;
   }
```
arcoraven, line 33: What's the purpose of this try/catch + custom error type, vs just letting the code throw?
```typescript
const kmsWalletParams = await fetchAwsKmsWalletParams(overrides);
// This may throw to the caller
```

src/tests/shared/client.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const TEST_CLIENT = createThirdwebClient(
+  secretKey
+    ? { secretKey }
+    : {
+        clientId: "TEST",
+        // if we don't have a secret key, we can use a public gateway for testing?
+        config: { storage: { gatewayUrl: "https://cf-ipfs.com" } },
+      },
+);
```
arcoraven, line 13: Test should just fail if no secret key. We already require one to run Engine.

src/server/utils/wallets/createAwsKmsWallet.ts
d4mr: create backend wallet endpoint uses the `createAwsKmsWallet` method. By having a custom error class it can then know for sure error is 400

src/server/routes/configuration/wallets/update.ts
```diff
@@ -10,16 +10,11 @@ import { responseBodySchema } from "./get";
 
 const requestBodySchema = Type.Union([
   Type.Object({
-    type: Type.Literal(WalletType.local),
-  }),
-  Type.Object({
-    type: Type.Literal(WalletType.awsKms),
```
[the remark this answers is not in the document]
d4mr: users might want to update for multiple wallet types at once. Eg: I want to pass in both GCP and AWS configuration in the same request, no reason not to allow that

src/server/utils/wallets/createAwsKmsWallet.ts
d4mr: the full version of this is to have an EngineServiceError that also contains more metadata, and each service can extend this. That would be a new PR though

src/server/utils/wallets/awsKmsArn.ts
```diff
@@ -4,11 +4,13 @@
 export function splitAwsKmsArn(arn: string) {
   // arn:aws:kms:<region>:<account-id>:key/<key-id>
   const parts = arn.split(":");
-  const keyId = parts[parts.length - 1].split("/")[1];
+  if (!parts[5]) throw new Error("Invalid AWS KMS ARN");
```
arcoraven, line 7: Nit: a length check is more intuitive to me

src/tests/gcp-resource-path.test.ts
```diff
@@ -0,0 +1,84 @@
+import { describe, expect, it } from "vitest";
+import {
+  getGcpKmsResourcePath,
+  splitGcpKmsResourcePath,
+} from "../server/utils/wallets/gcpKmsResourcePath"; // Replace with the actual file name
```
arcoraven, line 5: remove comment

arcoraven (approved):

PR #628, opened 2024-08-28
feat: nonce stats endpoint + queue/nonce log cron

d4mr:
## PR-Codex overview
This PR focuses on adding a `NonceHealthCheckQueue` class, worker, and related functionality for monitoring nonce health in the application.

### Detailed summary
- Added `NonceHealthCheckQueue` class with BullMQ integration
- Implemented `initNonceHealthCheckWorker` for nonce health monitoring
- Added functions to check queue status and update nonce history
- Updated logger with custom log levels and filters

> The following files were skipped due to too many changes: `src/worker/tasks/nonceHealthCheckWorker.ts`, `src/server/routes/admin/nonces.ts`


d4mr:
closing in favour of #628

d4mr:
wait nvm this is #628

src/server/routes/admin/nonces.ts
```diff
@@ -0,0 +1,169 @@
...
+import { walletWithAddressParamSchema } from "../../schemas/wallet";
+
+export const responseBodySchema = Type.Object({
+  result: Type.Array(
+    Type.Object({
+      walletAddress: Type.String({
+        description: "Backend Wallet Address",
+        examples: ["0xcedf3b4d8f7f1f7e0f7f0f7f0f7f0f7f0f7f0f7f"],
+      }),
+      onchainNonce: Type.String({
```
arcoraven, line 25: We should be consistent representing nonces everywhere. Besides the fact that Redis requires strings, IMO I like to use numbers everywhere. WDYT? e.g. this would be a `Type.Integer()`

Or better yet a common type we'd reuse:
```typescript
export const WalletNonceSchema = Type.Integer({
  description: "A wallet nonce",
  examples: [123],
})

// Usage
onchainNonce: {
  ...WalletNonceSchema,
  description: "Last mined nonce"
}
```

src/server/routes/admin/nonces.ts
```diff
@@ -0,0 +1,169 @@
...
+        description: "Last mined nonce",
+        examples: ["0"],
+      }),
+      lastUsedNonce: Type.String({
```
arcoraven, line 29: Nit: "lastUsed" is a bit ambiguous. Maybe:
- lastAcquiredNonce?
- dbNonce?
- engineNonce? (as opposed to onchainNonce)

src/server/routes/admin/nonces.ts
```diff
@@ -0,0 +1,169 @@
...
+          examples: ["0"],
+        }),
+        {
+          description:
+            "Nonces that were acquired but failed to be sent to the blockchain, waiting to be recycled or cancelled.",
+        },
+      ),
+      chainId: Type.Number({
+        description: "Chain ID",
+        examples: [80001],
```
arcoraven, line 53: Nit: use 80002 as the example since 80001 is deprecated

src/server/routes/admin/nonces.ts
```diff
@@ -0,0 +1,169 @@
...
+      operationId: "transactionDetails",
+      querystring: walletWithAddressParamSchema,
+      response: {
+        ...standardResponseSchema,
+        [StatusCodes.OK]: responseBodySchema,
+      },
+      hide: true,
+    },
+    handler: async (request, reply) => {
+      const { walletAddress, chain } = request.query;
```
arcoraven, line 91: Nit in preparation for the future: We could avoid `walletWithAddressParamSchema` and make a new type that accepts the input var `chainId` instead of `chain`

src/server/routes/admin/nonces.ts
```diff
@@ -0,0 +1,169 @@
...
+} = {}) => {
+  const lastUsedNonceKeys = await getLastUsedNonceKeys(walletAddress, chainId);
+
+  const result = await Promise.all(
+    lastUsedNonceKeys.map(async (key) => {
+      const { chainId, walletAddress } = splitLastUsedNonceKey(key);
+
+      const [lastUsedNonce, sentNonces, recycledNonces, onchainNonce] =
+        await Promise.all([
+          redis.get(lastUsedNonceKey(chainId, walletAddress)),
```
arcoraven, line 116: Nit: there's a `inspectNonce()` function that does this

src/server/routes/admin/nonces.ts
```diff
@@ -0,0 +1,169 @@
...
+    client: thirdwebClient,
+    chain: await getChain(chainId),
+  });
+
+  // The next unused nonce = transactionCount.
+  const transactionCount = await eth_getTransactionCount(rpcRequest, {
+    address: walletAddress,
+  });
+
+  const onchainNonce = transactionCount + 1;
```
arcoraven, line 167: You want `transactionCount - 1` here I believe.

The naming of `onchain nonce` is a bit ambiguous. Do you want:
1. The _next unused_ nonce?
2. The _last used_ nonce?

I think you probably want option 2 since it matches the db nonce we're querying (the last used db nonce).

src/worker/tasks/nonceHealthCheckWorker.ts
```diff
@@ -0,0 +1,176 @@
...
+  if (historicalStates.length < CHECK_PERIODS) return false;
+
+  const oldestOnchainNonce = historicalStates[CHECK_PERIODS - 1].onchainNonce;
+
+  // if for every period, the onchain nonce has not changes, and the internal nonce has strictly increased
+  // then the queue is stuck
+  const isStuckForAllPeriods = historicalStates.every((state, index) => {
+    if (index === historicalStates.length - 1) return true; // Last (oldest) state
+
+    const nextState = historicalStates[index + 1];
```
arcoraven, line 101: Nit: Since states are ordered new -> old, `nextState` is misleading since it's the actually the previous state.

Suggestion: Either `getHistoricalNonceStates()` should return sorted old -> new (use RPUSH instead of LPUSH) or rename this `prevState`

src/worker/tasks/nonceHealthCheckWorker.ts
```diff
@@ -0,0 +1,176 @@
...
+    return (
+      state.onchainNonce === oldestOnchainNonce &&
```
arcoraven, line 103: Nit: It's the same thing but makes it a bit simpler to read IMO to make this line `state.onchainNonce === prevState.onchainNonce`. Since in order for the queue to be marked as "stuck", the onchainNonce value must be the same for all periods. Then you can remove line 94.

src/worker/tasks/nonceHealthCheckWorker.ts
```diff
@@ -0,0 +1,176 @@
...
+    const message =
+      `[WALLET_HEALTH] ${result.walletAddress}:${result.chainId} ` +
+      `isStuck:${result.isStuck} ` +
+      `onchainNonce:${result.onchainNonce} ` +
+      `internalNonce:${result.internalNonce}`;
+
+    if (result.isStuck) {
+      logger({
+        service: "worker",
+        level: "warn",
```
arcoraven, line 167: IMO if a queue is stuck the log level should be something where we would monitor and likely page on-call. I'd make this level `fatal` or `error`.

No need to change the actual message. We can create monitors on the level, so the if statement can just control the level:
```typescript
logger({
  service: "worker",
  level: result.isStuck ? "fatal" : "info",
  message,
})
```

src/worker/tasks/nonceHealthCheckWorker.ts
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// Get historical nonce states
+async function getHistoricalNonceStates(
+  walletAddress: Address,
+  chainId: number,
+  periods: number,
+): Promise<NonceState[]> {
+  const key = nonceHistoryKey(walletAddress, chainId);
+  const historicalStates = await redis.lrange(key, 0, periods - 1);
```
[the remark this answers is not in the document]
d4mr: handling for the edge case when CHECK_PERIODS changes between deployments

src/server/routes/admin/nonces.ts
d4mr: not sure how to use this with pipeline

src/server/routes/admin/nonces.ts
d4mr: I don't like the idea of having both schemas hang around, would make things confusing... do we namespace the newer one with "v3" or similar? and later remove the namespace when we actually move to v3?

src/server/routes/admin/nonces.ts
d4mr: agreed, it's just the redis terminology + it's an admin route so wanted to keep it consistent between the two
would prefer a better name if breaking consistency between redis <> response is okay

src/server/routes/admin/nonces.ts
```diff
@@ -0,0 +1,192 @@
...
+      pipelineResults.slice(pipelineOffset, pipelineOffset + 3);
+
+    return {
+      walletAddress: key.walletAddress,
+      chainId: key.chainId,
+      onchainNonce: onchainNonces[index],
+      lastUsedNonce: parseInt(lastUsedNonceResult[1] as string) ?? 0,
+      sentNonces: (sentNoncesResult[1] as string[])
+        .map((nonce) => parseInt(nonce))
+        .sort((a, b) => b - a),
```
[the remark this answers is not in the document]
d4mr: I know you mentioned clarity >>> minor perf gains, but this one is painful lol. What about somewhere in the middle, a helper:
` const descendingSort = (a, b) => (b - a)`
and usage: 
` .sort(descendingSort)`

src/server/routes/admin/nonces.ts
```diff
@@ -0,0 +1,192 @@
...
+}
+
+export const getNonceDetails = async ({
+  walletAddress,
+  chainId,
+}: {
+  walletAddress?: string;
+  chainId?: string;
+} = {}) => {
+  const lastUsedNonceKeys = await getLastUsedNonceKeys(walletAddress, chainId);
```
arcoraven, line 103: Can we use `getUsedBackendWallets()` here? I think `getLastUsedNonceKeys()` doesn't need to exist anymore

src/worker/tasks/nonceHealthCheckWorker.ts
```diff
@@ -0,0 +1,165 @@
...
+  walletAddress: Address,
+  chainId: number,
+): Promise<NonceState> {
+  const [onchainNonce, largestSentNonce] = await Promise.all([
+    getLastUsedOnchainNonce(chainId, walletAddress),
+    inspectNonce(chainId, walletAddress),
+  ]);
+
+  return {
+    onchainNonce: parseInt(onchainNonce.toString()),
```
arcoraven, line 116: Isn't `onchainNonce` is already a number? This line stringifies it, then parses int again?

arcoraven (approved):

d4mr:
tested, needed some logger changes:
image

src/worker/queues/nonceHealthCheckQueue.ts
```diff
@@ -0,0 +1,14 @@
+import { Queue } from "bullmq";
+import { redis } from "../../utils/redis/redis";
+import { defaultJobOptions } from "./queues";
+
+export class NonceHealthCheckQueue {
+  static q = new Queue<string>("nonceHealthCheck", {
```
arcoraven, line 6: nit: The other queue names are kebab-cased. Could you make this kebab-case too? (There was no specific reason other than I just started them that way)

src/worker/tasks/nonceHealthCheckWorker.ts
```diff
@@ -0,0 +1,165 @@
...
+import { logWorkerExceptions } from "../queues/queues";
+
+// Configuration
+
+// Number of consecutive periods to check
+// Checking over multiple periods helps to avoid false positives due to timing issues
+const CHECK_PERIODS = 3;
+
+// Frequency of the worker
+const RUN_FREQUENCY_SECONDS = 20; // Run every 20 seconds
```
arcoraven, line 20: This used to be every minute? I think every minute still makes sense unless you had a reason to make it run more frequently.

src/worker/tasks/nonceHealthCheckWorker.ts
arcoraven: bump on this comment

src/utils/logger.ts
```diff
@@ -1,31 +1,57 @@
...
 // Custom filter for stdout transport
-const filterOnlyInfoAndWarn = format((info) => {
-  if (info.level === "error") {
-    return false; // Exclude 'error' level logs
+const filterNonErrors = format((info) => {
+  if (info.level !== "error" && info.level !== "fatal") {
+    return info; // only include non-error logs
   }
-  return info;
+  return false;
```
arcoraven, line 34: This logic is correct but seems like it can be more readable:

Instead of:
```
if not error and not fatal
  return info
else
  return false
```

This seems like the same but more readable:
```
if error or fatal
  return false
else
  return info
```

arcoraven (approved):

PR #639, opened 2024-09-03
feat: Add redis caching for account factory address

d4mr:
## PR-Codex overview
This PR focuses on enhancing the `generateSignedUserOperation` function in `userOperation.ts` by introducing a caching mechanism to retrieve the Factory Contract Address more efficiently.

### Detailed summary
- Added `getAddress` function
- Introduced caching for Factory Contract Address retrieval
- Improved error handling for missing Factory Contract Address

### DIFF
--- a/src/utils/transaction/userOperation.ts
+++ b/src/utils/transaction/userOperation.ts
@@ -1,6 +1,7 @@
 import {
   Address,
   Hex,
+  getAddress,
   getContract,
   prepareContractCall,
   readContract,
@@ -13,6 +14,7 @@ import {
 } from "thirdweb/wallets/smart";
 import { getAccount } from "../account";
 import { getChain } from "../chain";
+import { redis } from "../redis/redis";
 import { thirdwebClient } from "../sdk";
 import { QueuedTransaction } from "./types";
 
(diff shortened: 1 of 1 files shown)


jakeloo (approved):

src/utils/transaction/userOperation.ts
```diff
@@ -43,12 +45,31 @@ export const generateSignedUserOperation = async (
     address: accountAddress as Address,
   });
 
+  let accountFactoryAddress: Address | undefined;
+
+  try {
+    const cachedFactoryAddress = await redis.get(
+      `account-factory:${accountAddress.toLowerCase()}`,
+    );
+    accountFactoryAddress = getAddress(cachedFactoryAddress ?? "");
```
arcoraven, line 54: This is indirect to what you want:
- if cachedFactoryAddress exists, set accountFactoryAddress to the checksum version of it
- else accountFactoryAddress is undefined or null

Write the code that way. The current way has some unexpected behavior (what does `getAddress("")` do? You have to read the code to find out):

```typescript
  let accountFactoryAddress: Address | undefined;

  try {
    const cached = await redis.get(
      `account-factory:${accountAddress.toLowerCase()}`,
    );
    if (cached) {
      accountFactoryAddress = getAddress(cached);
    }
  } catch {}
```

src/utils/transaction/userOperation.ts
```diff
@@ -43,12 +45,31 @@ export const generateSignedUserOperation = async (
...
+
   // Resolve Factory Contract Address from Smart-Account Contract
-  const accountFactoryAddress = await readContract({
-    contract: smartAccountContract,
-    method: "function factory() view returns (address)",
-    params: [],
-  });
+
+  if (!accountFactoryAddress) {
+    const onchainFactoryResult = (await readContract({
```
arcoraven, line 60: Unless I'm missing something, shouldn't you just set the response directly to `accountFactoryAddress` instead of declaring a new var?

src/utils/transaction/userOperation.ts
```diff
@@ -43,12 +45,31 @@ export const generateSignedUserOperation = async (
...
+      contract: smartAccountContract,
+      method: "function factory() view returns (address)",
+      params: [],
+    })) as Address;
+    accountFactoryAddress = onchainFactoryResult;
```
arcoraven, line 65: You're missing a line here to set the cache after retrieving it.

PR #645, opened 2024-09-04
add `x-account-factory-address` header to contract-write and send-transaction

d4mr:
## PR-Codex overview
The focus of this PR is to add support for an `accountFactoryAddress` parameter in transactions and user operations.

### Detailed summary
- Added `accountFactoryAddress` property to transaction and user operation types
- Implemented `createBadAddressError` function for invalid address errors
- Updated transaction queueing and contract writing to handle `accountFactoryAddress`
- Added `maybeAddress` helper function to parse and validate addresses

### DIFF
--- a/src/server/routes/contract/write/write.ts
+++ b/src/server/routes/contract/write/write.ts
@@ -95,9 +95,10 @@ export async function writeToContract(fastify: FastifyInstance) {
         extension: "none",
         idempotencyKey,
         txOverrides,
-        accountFactoryAddress: accountFactoryAddress
-          ? maybeAddress(accountFactoryAddress)
-          : undefined,
+        accountFactoryAddress: maybeAddress(
+          accountFactoryAddress,
+          "x-account-factory-address",
+        ),
       });
 
       reply.status(StatusCodes.OK).send({
--- a/src/server/routes/contract/write/write.ts
+++ b/src/server/routes/contract/write/write.ts
@@ -11,7 +11,10 @@ import {
   transactionWritesResponseSchema,
 } from "../../../schemas/sharedApiSchemas";
 import { txOverridesWithValueSchema } from "../../../schemas/txOverrides";
-import { walletWithAAHeaderSchema } from "../../../schemas/wallet";
+import {
+  maybeAddress,
+  walletWithAAHeaderSchema,
+} from "../../../schemas/wallet";
 import { getChainIdFromChain } from "../../../utils/chain";
 
 // INPUT
@@ -66,6 +69,7 @@ export async function writeToContract(fastify: FastifyInstance) {
         "x-backend-wallet-address": walletAddress,
         "x-account-address": accountAddress,
         "x-idempotency-key": idempotencyKey,
+        "x-account-factory-address": accountFactoryAddress,
       } = request.headers as Static<typeof walletWithAAHeaderSchema>;
 
       const chainId = await getChainIdFromChain(chain);
@@ -76,6 +80,7 @@ export async function writeToContract(fastify: FastifyInstance) {
         accountAddress,
         abi,
       });
+
       const tx = contract.prepare(functionName, args, {
         value: txOverrides?.value,
         gasLimit: txOverrides?.gas,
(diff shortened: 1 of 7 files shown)


src/server/schemas/wallet/index.ts
```diff
@@ -20,8 +21,25 @@ export const walletWithAAHeaderSchema = Type.Object({
...
 });
 
+/*
+ * Helper function to parse an address string.
+ * Returns undefined if the address is invalid.
+ */
+export function maybeAddress(address: string | undefined) {
+  if (!address) return undefined;
+  try {
+    return getAddress(address);
```
arcoraven, line 37: Any reason to silently ignore invalid addresses? I think if an invalid address is passed, the expectation should be to error instead of assume no address. That way the developer can fix it instead of wondering why the tx looks like it's going through but not working.

src/server/routes/contract/write/write.ts
```diff
@@ -90,6 +95,9 @@ export async function writeToContract(fastify: FastifyInstance) {
         extension: "none",
         idempotencyKey,
         txOverrides,
+        accountFactoryAddress: accountFactoryAddress
```
arcoraven, line 98: `maybeAddress()` already accepts undefined so the ternary is unneeded here

src/server/schemas/wallet/index.ts
```diff
@@ -20,8 +21,25 @@ export const walletWithAAHeaderSchema = Type.Object({
...
+    ...AddressSchema,
+    description: "Smart account factory address",
+  }),
 });
 
+/*
+ * Helper function to parse an address string.
+ * Returns undefined if the address is invalid.
+ */
+export function maybeAddress(address: string | undefined) {
```
arcoraven, line 34: IMO this function belongs in the primitiveTypes.ts utils file which contain pure functions that do common things like this.

src/utils/transaction/userOperation.ts
```diff
@@ -43,12 +45,26 @@ export const generateSignedUserOperation = async (
...
+      });
+
+      accountFactoryAddress = getAddress(onchainAccountFactoryAddress);
+    } catch (e) {
+      // if no factory address is found, throw an error
+      throw new Error(
+        `Failed to find factory address for account '${accountAddress}' on chain '${chainId}'`,
+      );
+    }
+  }
```
arcoraven, line 67: Not sure if this `catch` is necessary unless the caught exception is hard to parse. Usually I try to avoid catching + rethrowing unless you explicitly want to wrap the error message.

src/server/schemas/wallet/index.ts
```diff
@@ -20,8 +21,25 @@ export const walletWithAAHeaderSchema = Type.Object({
     ...AddressSchema,
     description: "Smart account address",
   }),
+  "x-account-factory-address": Type.Optional({
+    ...AddressSchema,
+    description: "Smart account factory address",
```
arcoraven, line 26: Nit: Leave a note that if omitted, Engine will try to auto-detect it.

arcoraven (approved):

src/server/schemas/wallet/index.ts
d4mr: we already have `getAddress` that errors out. I built the interface such that if an incorrect address is passed into `x-account-factory-address`, it will ignore the header. This felt necessary for v1 for backwards compatibility reasons. Now we can make it fail loudly too. Changing it 👍

src/server/schemas/wallet/index.ts
d4mr: ty, wasn't sure of where to put it

src/utils/transaction/userOperation.ts
d4mr: I wanted us to explicitly know from logs that get `factoryAddress` step went wrong, otherwise it's a read contract call error

src/server/schemas/wallet/index.ts
d4mr: now that it throws a custom error, not sure if it belongs there anymore?

arcoraven (approved):

PR #647, opened 2024-09-04
Added prometheus metrics for response, queued, sent, mined, nonces

d4mr:
## PR-Codex overview
This PR adds Prometheus metrics support to the application. It introduces metrics recording for queued, mined, and sent transactions, along with a middleware for recording response metrics.

### Detailed summary
- Added `prom-client` for Prometheus metrics
- Added metrics recording for queued, mined, and sent transactions
- Implemented middleware for recording response metrics

> The following files were skipped due to too many changes: `yarn.lock`, `src/utils/prometheus.ts`

### DIFF
--- a/src/server/middleware/prometheus.ts
+++ b/src/server/middleware/prometheus.ts
@@ -1,7 +1,16 @@
-import { FastifyInstance, FastifyReply, FastifyRequest } from "fastify";
+import fastify, {
+  FastifyInstance,
+  FastifyReply,
+  FastifyRequest,
+} from "fastify";
+import { env } from "../../utils/env";
 import { enginePromRegister, recordMetrics } from "../../utils/prometheus";
 
 export const withPrometheus = async (server: FastifyInstance) => {
+  if (!env.METRICS_ENABLED) {
+    return;
+  }
+
   server.addHook(
     "onResponse",
     async (req: FastifyRequest, res: FastifyReply) => {
@@ -21,10 +30,20 @@ export const withPrometheus = async (server: FastifyInstance) => {
       });
     },
   );
-
   // Expose metrics endpoint
-  server.get("/metrics", async (request, reply) => {
+
+  const metricsServer = fastify({
+    disableRequestLogging: true,
+  });
+
+  metricsServer.get("/metrics", async (request, reply) => {
     reply.header("Content-Type", enginePromRegister.contentType);
     return enginePromRegister.metrics();
+  });
+
+  await metricsServer.ready();
+  metricsServer.listen({
+    host: env.METRICS_HOST ?? env.HOST,
+    port: env.METRICS_PORT,
   });
 };
--- a/src/server/middleware/prometheus.ts
+++ b/src/server/middleware/prometheus.ts
@@ -1,0 +1,23 @@
+import { FastifyInstance, FastifyReply, FastifyRequest } from "fastify";
+import { recordMetrics } from "../../utils/prometheus";
+
+export const withPrometheus = async (server: FastifyInstance) => {
+  server.addHook(
+    "onResponse",
+    async (req: FastifyRequest, res: FastifyReply) => {
+      const { method, url, ip, routeOptions } = req;
+      const { statusCode } = res;
+      const duration = res.elapsedTime; // Use Fastify's built-in timing
+
+      // Record metrics
+      recordMetrics({
+        event: "response_sent",
+        params: {
+          endpoint: new URL(url).pathname,
+          statusCode,
+          duration: duration / 1000, // Convert to seconds for Prometheus
+        },
+      });
+    },
+  );
+};
--- a/src/server/middleware/prometheus.ts
+++ b/src/server/middleware/prometheus.ts
@@ -4,11 +4,9 @@ import { recordMetrics } from "../../utils/prometheus";
 
 export const withPrometheus = async (server: FastifyInstance) => {
   if (!env.METRICS_ENABLED) {
-    console.log("Metrics are disabled");
     return;
   }
 
-  console.log("Metrics are enabled");
   server.addHook(
     "onResponse",
     (req: FastifyRequest, res: FastifyReply, done) => {
--- a/src/server/middleware/prometheus.ts
+++ b/src/server/middleware/prometheus.ts
@@ -1,6 +1,5 @@
 import { FastifyInstance, FastifyReply, FastifyRequest } from "fastify";
-import { register } from "prom-client";
-import { recordMetrics } from "../../utils/prometheus";
+import { enginePromRegister, recordMetrics } from "../../utils/prometheus";
 
 export const withPrometheus = async (server: FastifyInstance) => {
   server.addHook(
@@ -25,7 +24,7 @@ export const withPrometheus = async (server: FastifyInstance) => {
 
   // Expose metrics endpoint
   server.get("/metrics", async (request, reply) => {
-    reply.header("Content-Type", register.contentType);
-    return register.metrics();
+    reply.header("Content-Type", enginePromRegister.contentType);
+    return enginePromRegister.metrics();
   });
 };
--- a/src/server/middleware/prometheus.ts
+++ b/src/server/middleware/prometheus.ts
@@ -1,10 +1,6 @@
-import fastify, {
-  FastifyInstance,
-  FastifyReply,
-  FastifyRequest,
-} from "fastify";
+import { FastifyInstance, FastifyReply, FastifyRequest } from "fastify";
 import { env } from "../../utils/env";
-import { enginePromRegister, recordMetrics } from "../../utils/prometheus";
+import { recordMetrics } from "../../utils/prometheus";
 
 export const withPrometheus = async (server: FastifyInstance) => {
   if (!env.METRICS_ENABLED) {
@@ -35,20 +31,4 @@ export const withPrometheus = async (server: FastifyInstance) => {
       done();
     },
   );
-  // Expose metrics endpoint
-
-  const metricsServer = fastify({
-    disableRequestLogging: true,
-  });
-
-  metricsServer.get("/metrics", async (request, reply) => {
-    reply.header("Content-Type", enginePromRegister.contentType);
-    return enginePromRegister.metrics();
-  });
-
-  await metricsServer.ready();
-  metricsServer.listen({
-    host: env.METRICS_HOST ?? env.HOST,
-    port: env.METRICS_PORT,
-  });
 };
--- a/src/server/middleware/prometheus.ts
+++ b/src/server/middleware/prometheus.ts
@@ -8,13 +8,16 @@ import { enginePromRegister, recordMetrics } from "../../utils/prometheus";
 
 export const withPrometheus = async (server: FastifyInstance) => {
   if (!env.METRICS_ENABLED) {
+    console.log("Metrics are disabled");
     return;
   }
 
+  console.log("Metrics are enabled");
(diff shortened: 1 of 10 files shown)


src/utils/env.ts
```diff
@@ -79,6 +79,7 @@ export const env = createEnv({
       .default("default"),
     GLOBAL_RATE_LIMIT_PER_MIN: z.coerce.number().default(400 * 60),
     DD_TRACER_ACTIVATED: z.coerce.boolean().default(false),
+    PROMETHEUS_PUSHGATEWAY_URL: z.string().url().optional(),
```
d4mr, line 82: ah need to remove this now that we do pull

src/utils/prometheus.ts
```diff
@@ -0,0 +1,183 @@
+import { Counter, Gauge, Histogram, Registry } from "prom-client";
+import { getUsedBackendWallets, inspectNonce } from "../db/wallets/walletNonce";
+import { getLastUsedOnchainNonce } from "../server/routes/admin/nonces";
+
+const nonceMetrics = new Gauge({
+  name: "engine_nonces",
+  help: "Current nonce values and health for backend wallets",
+  labelNames: ["wallet", "chain", "type"],
```
jakeloo, line 8: `chain` -> `chain_id` to be consistent throughout?

jakeloo: `wallet` -> `wallet_address` to be consistent

src/utils/prometheus.ts
```diff
@@ -0,0 +1,183 @@
...
+      const [onchainNonce, engineNonce] = await Promise.all([
+        getLastUsedOnchainNonce(chainId, walletAddress),
+        inspectNonce(chainId, walletAddress),
+      ]);
+
+      this.set(
+        {
+          wallet: walletAddress,
+          chain: chainId.toString(),
+          type: "onchain",
```
jakeloo, line 22: instinctively, i thought `type` = wallet type

src/worker/tasks/sendTransactionWorker.ts
```diff
@@ -106,9 +108,27 @@ const handler: Processor<any, void, string> = async (job: Job<string>) => {
...
+          },
+        });
       }
     } else if (resultTransaction.status === "errored") {
       _reportUsageError(resultTransaction);
+      recordMetrics({
+        event: "transaction_sent",
+        params: {
+          chainId: transaction.chainId.toString(),
+          success: true,
```
jakeloo, line 127: error?

src/worker/tasks/sendTransactionWorker.ts
```diff
@@ -106,9 +108,27 @@ const handler: Processor<any, void, string> = async (job: Job<string>) => {
       // Report usage only on the first transaction send.
       if (transaction.status === "queued") {
```
jakeloo, line 109: interesting that this is not `resultTransaction`?

src/utils/prometheus.ts
```diff
@@ -0,0 +1,183 @@
...
+const transactionsMinedTotal = new Counter({
+  name: "engine_transactions_mined_total",
+  help: "Total number of transactions mined",
+  labelNames: ["chain_id", "wallet_address"],
+  registers: [register],
+});
+
+// Transaction duration histograms
+const queueToSentDuration = new Histogram({
+  name: "engine_transaction_queue_to_sent_seconds",
```
jakeloo, line 119: we should pick `engine_transactions_` or `engine_transaction_` 
singular or plural, and be consistent

d4mr: traefik does this:
for histograms: `traefik_service_request_duration_seconds_sum`
for counters: `traefik_service_requests_total`

src/worker/tasks/sendTransactionWorker.ts
d4mr: `transaction.status: "queued" | "sent"`
`resultTransaction.status: "sent" | "errored"`

src/server/middleware/prometheus.ts
```diff
@@ -0,0 +1,54 @@
...
+  FastifyInstance,
+  FastifyReply,
+  FastifyRequest,
+} from "fastify";
+import { env } from "../../utils/env";
+import { enginePromRegister, recordMetrics } from "../../utils/prometheus";
+
+export const withPrometheus = async (server: FastifyInstance) => {
+  if (!env.METRICS_ENABLED) {
+    console.log("Metrics are disabled");
```
arcoraven, line 11: Let's use the `logger` helper so all logging is consolidated and reuses the same pattern/libraries.

src/server/middleware/prometheus.ts
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  metricsServer.get("/metrics", async (request, reply) => {
+    reply.header("Content-Type", enginePromRegister.contentType);
+    return enginePromRegister.metrics();
+  });
+
+  await metricsServer.ready();
+  metricsServer.listen({
+    host: env.METRICS_HOST ?? env.HOST,
```
arcoraven, line 51: In what situation should this be overridden? 
From the [`listen()` docs](https://fastify.dev/docs/latest/Reference/Server/#listen) you'd only override it to support IPv4/v6 addresses. I'm not really sure of the behavior we'd want to allow, so I wonder if we should just default to `env.HOST` since we want this metrics server to probably just listen on the same IP types?

tl;dr: Should we remove `METRICS_HOST` as an env var?

src/utils/prometheus.ts
```diff
@@ -0,0 +1,185 @@
...
+// Define strict types for each event
+type ResponseSentParams = {
+  endpoint: string;
+  statusCode: string;
+  method: string;
+  duration: number;
+};
+
+type TransactionQueuedParams = {
+  chainId: string;
```
arcoraven, line 51: Why not number?

src/utils/prometheus.ts
```diff
@@ -0,0 +1,185 @@
...
+  labelNames: ["chain_id", "wallet_address"],
+  registers: [enginePromRegister],
+});
+
+// Transaction duration histograms
+const queueToSentDuration = new Histogram({
+  name: "engine_transaction_queue_to_sent_seconds",
+  help: "Duration from queue to sent in seconds",
+  labelNames: ["chain_id", "wallet_address"],
+  buckets: [1, 5, 15, 30, 60, 120, 300, 600],
```
arcoraven, line 122: Unfortunate that timer data in Prometheus is so clunky. I wonder if having different bucket sizes complicates viewing the data.

I'm thinking the "reasonable ranges" for the following are:
- queued to sent: 50ms to 12 hours. The widest range since a backed up queue (or stuck nonce) could mean txs are in queue for hours.
- sent to mined: 200ms to 30 minutes

src/utils/prometheus.ts
```diff
@@ -0,0 +1,185 @@
...
+
+const sentToMinedDuration = new Histogram({
+  name: "engine_transaction_sent_to_mined_seconds",
+  help: "Duration from sent to mined in seconds",
+  labelNames: ["chain_id", "wallet_address"],
+  buckets: [10, 30, 60, 120, 300, 600, 1200, 1800],
+  registers: [enginePromRegister],
+});
+
+const endToEndDuration = new Histogram({
```
arcoraven, line 134: Nit: `queuedToMined` is clearer to me since it's more explicit and matches the pattern of the others.

Also not sure if this one's particularly useful since it's can be estimated as the sum of the other two.

src/utils/prometheus.ts
```diff
@@ -0,0 +1,185 @@
...
+  name: "engine_transaction_end_to_end_seconds",
+  help: "Duration from request to mined in seconds",
```
arcoraven, line 136: From queued to mined, since we aren't tracking "request to queued" (the logic in `mineTransactionWorker.ts` records `minedAt` - `queuedAt`)

src/utils/prometheus.ts
```diff
@@ -0,0 +1,185 @@
+import { Counter, Gauge, Histogram, Registry } from "prom-client";
+import { getUsedBackendWallets, inspectNonce } from "../db/wallets/walletNonce";
+import { getLastUsedOnchainNonce } from "../server/routes/admin/nonces";
+
+const nonceMetrics = new Gauge({
+  name: "engine_nonces",
+  help: "Current nonce values and health for backend wallets",
+  labelNames: ["wallet_address", "chain_id", "nonce_type"],
```
arcoraven, line 8: You can add `as const` to enforce Typescript ([source](https://github.com/siimon/prom-client?tab=readme-ov-file#strongly-typed-labels))

Also not sure on common syntax but the docs use camelCase for labels. It might clean up a lot of code to be more consistent and avoid bugs from using the wrong casing in the future.

src/utils/transaction/insertTransaction.ts
```diff
@@ -70,5 +72,13 @@ export const insertTransaction = async (
   });
   reportUsage([{ action: "queue_tx", input: queuedTransaction }]);
 
+  recordMetrics({
+    event: "transaction_queued",
+    params: {
+      chainId: insertedTransaction.chainId.toString(),
+      walletAddress: getAddress(insertedTransaction.from),
```
arcoraven, line 79: Nit: use `queuedTransaction` here since it's more accurate here (it's recording a "queued transaction" event). The `queuedTransaction.from` is already checksum case too.

src/utils/prometheus.ts
```diff
@@ -0,0 +1,185 @@
...
+};
+
+type TransactionSentParams = TransactionQueuedParams & {
+  success: boolean;
+  duration: number;
+};
+
+type TransactionMinedParams = TransactionQueuedParams & {
+  endToEndDuration: number;
+  duration: number;
```
arcoraven, line 62: Nit: for durations can we either include the units in the name or a TSDoc comment? It was unclear without digging a few layers deep.

Also comments would be helpful since `duration` and `endToEndDuration` sound similar (if we want to keep the latter).

src/utils/prometheus.ts
```diff
@@ -0,0 +1,185 @@
...
+});
+
+export const enginePromRegister = new Registry();
+
+enginePromRegister.registerMetric(nonceMetrics);
+
+// Define strict types for each event
+type ResponseSentParams = {
+  endpoint: string;
+  statusCode: string;
```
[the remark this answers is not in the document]
d4mr: number makes more sense in metrics when the `magnitude` of the number reflect something.
ie, higher or lower has real meaning. which is not the case for endpoint where it's a statusCode.

src/server/middleware/prometheus.ts
d4mr: this log line slipped out when debugging, do we even need it?

src/utils/prometheus.ts
d4mr: Agreed on the clunkiness. Histograms are our only option for aggregatable duration data in Prometheus' pull-based model. Unfortunate necessary trade-off to preserve information.

Changing the bucket sizes though 👍

src/utils/prometheus.ts
d4mr: nice! didn't know they supported it.
https://prometheus.io/docs/practices/naming/#labels Snake Case seems to be the standard

src/server/middleware/prometheus.ts
arcoraven: Not really, removing is fine too.

arcoraven (approved):

PR #649, opened 2024-09-06
chore: accept public key hash in kid header

arcoraven:
## PR-Codex overview
The focus of this PR is to update keypair retrieval by introducing a hash-based method and enhancing keypair caching for improved security and performance.

### Detailed summary
- Added `getKeypairByHash` function for hash-based keypair retrieval.
- Updated `getKeypairByPublicKey` to use hash internally.
- Modified `getKeypair` to handle both public key and hash.
- Refactored `handleKeypairAuth` to support both public key and hash.


src/server/middleware/auth.ts
```diff
@@ -165,18 +165,24 @@ export const onRequest = async ({
...
           return await handleAccessToken(jwt, req, getUser);
         } else if (publicKey === THIRDWEB_DASHBOARD_ISSUER) {
           return await handleDashboardAuth(jwt);
         } else {
-          return await handleKeypairAuth(jwt, req, publicKey);
+          return await handleKeypairAuth({ jwt, req, publicKey });
         }
       }
+
+      // Get the public key hash from the `kid` header.
```
arcoraven, line 181: Breaking change (though undocumented): The "kid" header accepts the hash instead of plaintext public key.

src/utils/cache/keypair.ts
```diff
@@ -1,20 +1,39 @@
...
 export const keypairCache = new Map<string, Keypairs | null>();
 
-export const getKeypair = async ({
-  publicKey,
-}: {
-  publicKey: string;
+/**
+ * Get a keypair by public key or hash.
+ */
+export const getKeypair = async (args: {
```
d4mr, line 10: nit: this could be made typesafe (I know we do runtime checks but still)

`type GetKeypairArgs = { publicKey: string } | {publicKeyHash: string}`

I understand current version is more convenient because it reduces upstream changes.

d4mr (approved):

PR #652, opened 2024-09-10
chore: toSerializableTransaction function for gas estimation

d4mr:
## PR-Codex overview
The focus of this PR is to refactor transaction handling logic by introducing a new function `getPopulatedOrErroredTransaction`.

### Detailed summary
- Added `toSerializableTransaction` import
- Refactored transaction preparation logic in `generateSignedUserOperation`
- Introduced `getPopulatedOrErroredTransaction` to handle transaction validation and population

### DIFF
--- a/src/worker/tasks/sendTransactionWorker.ts
+++ b/src/worker/tasks/sendTransactionWorker.ts
@@ -81,11 +81,48 @@ const handler: Processor<any, void, string> = async (job: Job<string>) => {
   // null = the transaction attemped to resend but was not needed. Ignore.
   // A thrown exception indicates a retry-able error occurred (e.g. RPC outage).
   let resultTransaction: SentTransaction | ErroredTransaction | null;
-  if (transaction.status === "queued") {
+
+  let populatedTransaction:
+    | Awaited<ReturnType<typeof toSerializableTransaction>>
+    | undefined;
+
+  const chain = await getChain(transaction.chainId);
+  // Populate the transaction to resolve gas values.
+  // This call throws if the execution would be reverted.
+  // The nonce is _not_ set yet.
+  try {
+    populatedTransaction = await toSerializableTransaction({
+      from: getChecksumAddress(transaction.from),
+      transaction: {
+        client: thirdwebClient,
+        chain,
+        ...transaction,
+        // Stub the nonce because it will be overridden later.
+        nonce: 1,
+      },
+    });
+  } catch (e: unknown) {
+    // If the transaction will revert, error.message contains the human-readable error.
+    const errorMessage = (e as Error)?.message ?? `${e}`;
+    job.log(`Failed to estimate gas: ${errorMessage}`);
+    resultTransaction = {
+      ...transaction,
+      status: "errored",
+      errorMessage,
+    };
+  }
+
+  // we know that if status is queued, populatedTransaction is defined
+  // typescript has a hard time understanding, but we can safely check that populatedTransaction is defined
+  if (transaction.status === "queued" && populatedTransaction) {
     if (transaction.isUserOp) {
       resultTransaction = await _sendUserOp(job, transaction);
     } else {
-      resultTransaction = await _sendTransaction(job, transaction);
+      resultTransaction = await _sendTransaction(
+        job,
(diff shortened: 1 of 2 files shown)


0xFirekeeper (approved):

d4mr:
@arcoraven should we try-catch-rethrow an easier to understand error?

arcoraven:
This will populate the transaction even on resends. The resend doesn't _use_ it though, so it's just an unneeded call. Maybe it's cleaner to just have a helper func `getPopulatedTransaction()` and call it from both `_sendUserOp()` and `_sendTransaction()` instead?

src/worker/tasks/sendTransactionWorker.ts
```diff
@@ -81,11 +81,48 @@ const handler: Processor<any, void, string> = async (job: Job<string>) => {
...
+        chain,
+        ...transaction,
+        // Stub the nonce because it will be overridden later.
+        nonce: 1,
+      },
+    });
+  } catch (e: unknown) {
+    // If the transaction will revert, error.message contains the human-readable error.
+    const errorMessage = (e as Error)?.message ?? `${e}`;
+    job.log(`Failed to estimate gas: ${errorMessage}`);
```
joaquim-verges, line 107: Nit: would change this msg even if it's internal to something like "tx validation failed" or something

src/worker/tasks/sendTransactionWorker.ts
```diff
@@ -81,11 +81,48 @@ const handler: Processor<any, void, string> = async (job: Job<string>) => {
...
   // A thrown exception indicates a retry-able error occurred (e.g. RPC outage).
   let resultTransaction: SentTransaction | ErroredTransaction | null;
-  if (transaction.status === "queued") {
+
+  let populatedTransaction:
+    | Awaited<ReturnType<typeof toSerializableTransaction>>
+    | undefined;
+
+  const chain = await getChain(transaction.chainId);
+  // Populate the transaction to resolve gas values.
```
joaquim-verges, line 90: Btw if gas value is passed here (make sure it's the 'gas' prop, not 'gasLimit' - then there will be no gas estimation and therefore no simulation.

I think that's actually what people expect in this case though, it's a way to 'force' the tx onchain. 

Lmk if that's fine

src/worker/tasks/sendTransactionWorker.ts
```diff
@@ -81,11 +81,48 @@ const handler: Processor<any, void, string> = async (job: Job<string>) => {
...
+      status: "errored",
+      errorMessage,
+    };
+  }
+
+  // we know that if status is queued, populatedTransaction is defined
+  // typescript has a hard time understanding, but we can safely check that populatedTransaction is defined
+  if (transaction.status === "queued" && populatedTransaction) {
     if (transaction.isUserOp) {
       resultTransaction = await _sendUserOp(job, transaction);
```
joaquim-verges, line 119: You need to pass the populated tx here though, you want to make sure you use the 'data' from the populated tx when you prep the execute call

src/worker/tasks/sendTransactionWorker.ts
d4mr: yep, this is what we expect

src/worker/tasks/sendTransactionWorker.ts
```diff
@@ -393,3 +381,43 @@ export const initSendTransactionWorker = () => {
...
+      ? queuedTransaction.accountAddress
+      : queuedTransaction.from;
+
+    if (!from) throw new Error("Invalid transaction parameters: from");
+
+    const populatedTransaction = await toSerializableTransaction({
+      from: getChecksumAddress(
+        queuedTransaction.isUserOp
+          ? queuedTransaction.accountAddress
+          : queuedTransaction.from,
```
joaquim-verges, line 402: could reuse the `from` variable here no?

joaquim-verges (approved):

src/worker/tasks/sendTransactionWorker.ts
d4mr: nice catch! thanks

PR #658, opened 2024-09-14
Feat: Add extraGas config, make cancels use higher gas

arcoraven:
## PR-Codex overview
The focus of this PR is to enhance the configuration handling in the server and worker modules.

### Detailed summary
- Added configuration schema and handling functions in `schemas/configuration.ts`
- Updated gas price calculations in `cancelTransaction.ts`
- Added routes for listing and setting configurations in `routes/index.ts`
- Implemented logic for listing and setting configurations in `routes/configuration/db/list.ts` and `routes/configuration/db/set.ts`
- Enhanced transaction processing in `sendTransactionWorker.ts` to consider extra gas configuration.

### DIFF
--- a/src/server/routes/configuration/db/list.ts
+++ b/src/server/routes/configuration/db/list.ts
@@ -1,0 +1,40 @@
+import { Type, type Static } from "@sinclair/typebox";
+import type { FastifyInstance } from "fastify";
+import { StatusCodes } from "http-status-codes";
+import { ConfigurationDB } from "../../../../db/configuration/db";
+import {
+  configurationSchema,
+  toConfigurationSchema,
+} from "../../../schemas/configuration";
+import { standardResponseSchema } from "../../../schemas/sharedApiSchemas";
+
+const responseBodySchema = Type.Object({
+  result: configurationSchema,
+});
+
+export async function listConfiguration(fastify: FastifyInstance) {
+  fastify.route<{
+    Reply: Static<typeof responseBodySchema>;
+  }>({
+    method: "GET",
+    url: "/configuration",
+    schema: {
+      summary: "List configuration",
+      description:
+        "List configuration values. Note: This does not include legacy configuration.",
+      tags: ["Configuration"],
+      operationId: "listConfiguration",
+      response: {
+        ...standardResponseSchema,
+        [StatusCodes.OK]: responseBodySchema,
+      },
+    },
+    handler: async (req, res) => {
+      const configuration = await ConfigurationDB.getAll();
+
+      res.status(200).send({
+        result: toConfigurationSchema(configuration),
+      });
+    },
+  });
+}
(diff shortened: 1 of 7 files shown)


src/server/routes/configuration/db/list.ts
```diff
@@ -0,0 +1,40 @@
...
+export async function listConfiguration(fastify: FastifyInstance) {
+  fastify.route<{
+    Reply: Static<typeof responseBodySchema>;
+  }>({
+    method: "GET",
+    url: "/configuration",
+    schema: {
+      summary: "List configuration",
+      description:
+        "List configuration values. Note: This does not include legacy configuration.",
```
joaquim-verges, line 24: i see this is the new way to add / fetch config flags?

seems a bit weird to me to use redis for it though no?

src/utils/transaction/cancelTransaction.ts
```diff
@@ -28,16 +28,16 @@ export const sendCancellationTransaction = async (
...
-    populatedTransaction.gasPrice *= 2n;
+    populatedTransaction.gasPrice *= 10n;
   }
   if (populatedTransaction.maxFeePerGas) {
-    populatedTransaction.maxFeePerGas *= 2n;
+    populatedTransaction.maxFeePerGas *= 10n;
   }
   if (populatedTransaction.maxFeePerGas) {
-    populatedTransaction.maxFeePerGas *= 2n;
+    populatedTransaction.maxFeePerGas *= 10n;
```
joaquim-verges, line 40: uuuh this looks wrong? you're 10x `maxFeePerGas` twice here

src/worker/tasks/sendTransactionWorker.ts
```diff
@@ -427,16 +433,25 @@ export const getPopulatedOrErroredTransaction = async (
...
+    }
+
+    const preparedTransaction = prepareTransaction({
+      client: thirdwebClient,
+      chain: await getChain(queuedTransaction.chainId),
+      ...queuedTransaction,
+      to: getChecksumAddress(to),
+      // if transaction is EOA, stub the nonce to reduce RPC calls
+      nonce: queuedTransaction.isUserOp ? undefined : 1,
+      extraGas,
```
joaquim-verges, line 449: i was kinda expecting this to be added per request, not as a config flag for the whole engine instance.

why isnt extraGas part of the txOverrides ?

arcoraven: Hmmm we could do that instead. Not sure if the need for this will commonly be for a chain or use case where the gas estimate is always too low, hence the global setting. 

A per-tx setting is more flexible but requires a code change on the dev side. If it's likely a dev would want to add gas only for certain requests, this makes more sense.

joaquim-verges: From experience this is on a per request basis. You only need this for calls that have some random execution or chain of delegate calls.

But doesn't make sense to add it for a basic transfer. 

Thinking about ecosystem engine instances that can be shared, def doesn't make sense to have this as a global flag

PR #702, opened 2024-10-03
Fix: Support cloudflare proxy IP

arcoraven:
Support cloudflare proxy IP only if trustProxy is enabled.

---

## PR-Codex overview
This PR focuses on improving logging messages, enhancing IP address checking functionality, and refactoring the `checkIpInAllowlist` function to return more detailed information about IP checks.

### Detailed summary
- Updated log message in `src/scripts/apply-migrations.ts` for migration completion.
- Added logging for `trustProxy` status in `src/server/index.ts`.
- Refactored `checkIpInAllowlist` in `src/server/middleware/auth.ts` to return an object containing `isAllowed` and `ip`.
- Updated error messages to include the checked IP instead of the request IP.

### DIFF
--- a/src/server/middleware/auth.ts
+++ b/src/server/middleware/auth.ts
@@ -408,12 +408,6 @@ const handleAccessToken = async (
       message: `Unauthorized IP address: ${req.ip}`,
     });
 
-    logger({
-      service: "server",
-      level: "error",
-      message: `[DEBUG] Request headers: ${JSON.stringify(req.headers)}`,
-    });
-
     return {
       isAuthed: false,
       error:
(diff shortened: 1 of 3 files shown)


src/server/middleware/auth.ts
```diff
@@ -523,12 +523,23 @@ const hashRequestBody = (req: FastifyRequest): string => {
...
+> => {
   const config = await getConfig();
-
   if (config.ipAllowlist.length === 0) {
-    return true;
+    return { isAllowed: true };
+  }
+
+  let ip = req.ip;
+  if (req.headers["cf-connecting-ip"]) {
```
d4mr, line 537: can we check for env.TRUST_PROXY or similar?
engine code is open source. An attacker can learn that we "blindly" trust this header and can circumvent IP allowlist for self-hosted engines

src/server/middleware/auth.ts
```diff
@@ -523,12 +523,23 @@ const hashRequestBody = (req: FastifyRequest): string => {
  * @returns boolean
  * @async
  */
-const checkIpInAllowlist = async (req: FastifyRequest) => {
+const checkIpInAllowlist = async (
+  req: FastifyRequest,
+): Promise<
+  { isAllowed: false; ip: string } | { isAllowed: true; ip?: string }
```
d4mr, line 529: why is `ip` optional in one of the response types?

d4mr (approved):

src/server/middleware/auth.ts
arcoraven: Because I'm not fetching it when there's no allowlist. But honestly it's just reading headers so I'll simplify it.

src/server/middleware/auth.ts
arcoraven: Good catch

PR #709, opened 2024-10-07
Smart Backend Wallets

d4mr:
## PR-Codex overview
This PR introduces enhancements to wallet management, including support for smart wallets, new wallet types, and updated database schemas. It also improves caching mechanisms and adds tests for new features.

### Detailed summary
- Added `entrypointAddress` and `accountFactoryAddress` to wallet schemas.
- Introduced smart wallet types: `smart:aws-kms`, `smart:gcp-kms`, and `smart:local`.
- Updated database migrations for new fields.
- Enhanced wallet creation functions to support smart wallets.
- Implemented caching for chain capabilities.
- Added tests for smart backend wallets.
- Deprecated some existing wallet functions and marked them for removal.

> The following files were skipped due to too many changes: `src/utils/cache/getWallet.ts`, `src/db/wallets/getWalletDetails.ts`, `src/utils/account.ts`


src/server/utils/wallets/getLocalWallet.ts
```diff
@@ -74,11 +78,18 @@ export const getLocalWallet = async ({
...
+    !walletDetails ||
+    walletDetails.type !== "local" ||
+    !walletDetails.encryptedJson
+  ) {
+    throw new Error(`Local Wallet not found for address ${walletAddress}`);
   }
 
+  const json = walletDetails.encryptedJson;
+
   const wallet = await Wallet.fromEncryptedJson(
```
d4mr, line 93: this can be moved over to using `decryptoJsonWallet` from `legacyLocalCrypto`. The interface isn't as clunky here though, so we can let it be too.

src/server/utils/wallets/createSmartWallet.ts
```diff
@@ -0,0 +1,152 @@
...
+
+const createSmartWallet = async ({
+  adminAccount,
+  accountFactoryAddress,
+}: {
+  adminAccount: Account;
+  accountFactoryAddress?: Address;
+}) => {
+  const smartAccount = smartWallet({
+    chain: defineChain(1),
```
joaquim-verges, line 27: I assume you're only using this to predict the smart wallet address?

The danger here is that it requires the account factory to be there on that chain.

I think for smart accounts it's better to recreate the smart wallet at runtime based on the chain coming in the request.

So you would create a normal backend wallet and just give it a smart flag in the db (which you kinda do already with the factory address) then on every request, compute the smart wallet based on the chain in the request. If factory not available on that chain = error.

src/server/utils/wallets/createLocalWallet.ts
```diff
@@ -1,23 +1,57 @@
...
 interface CreateLocalWallet {
   label?: string;
 }
 
-export const createLocalWallet = async ({
+/**
+ * Create a local wallet with a random private key
+ * Does not store the wallet in the database
+ */
+export const createLocalWallet = async () => {
```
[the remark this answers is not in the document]
arcoraven: Same comment about naming above. We're definitely overloading the word "create" and are going to trip ourselves up. This function generates one but doesn't store it anywhere. Maybe `generateLocalWallet()` more clearly conveys it is generating one in memory and there are no side effects.

src/server/utils/wallets/createSmartWallet.ts
d4mr: > I think for smart accounts it's better to recreate the smart wallet at runtime based on the chain coming in the request.

That's the intention here, did I miss something in the implementation?

> then on every request, compute the smart wallet based on the chain in the request
This would be very confusing. We must ensure smart wallet has same account address on every chain for UX. I assumed this was already always the case, so we are storing the predicted account address.

> If factory not available on that chain = error.
exactly

src/server/utils/wallets/legacyLocalCrypto.ts
```diff
@@ -0,0 +1,138 @@
...
+      p: number;
+      r: number;
+    };
+    mac: string;
+  };
+  id: string;
+  version: number;
+}
+
+export async function toEncryptedJson({
```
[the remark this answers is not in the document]
d4mr: this replaces the `LocalStorage` class implementation that we needed to do, which had dependencies on v4 SDK and ethers. We could alternatively use the ethers code directly, but I wanted to remove all ethers dependencies.

src/utils/account.ts
```diff
@@ -147,6 +147,33 @@ export const getAccount = async (args: {
       _accountsCache.set(cacheKey, account);
       return account;
     }
+    case WalletType.smartAwsKms:
+    case WalletType.smartGcpKms:
+    case WalletType.smartLocal: {
+      if (!walletDetails.accountSignerAddress) {
+        throw createCustomError(
```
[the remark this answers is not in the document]
d4mr: I saw line 38 so assumed it was okay to throw HTTP error here. Completely agree though, it's a bad pattern
```
  if (!walletDetails) {
    throw createCustomError(
      `No configured wallet found with address ${from}`,
      StatusCodes.BAD_REQUEST,
      "BAD_REQUEST",
    );
  }
  ```

File: src/server/utils/wallets/createSmartWallet.ts [hunk 6]
```diff
@@ -0,0 +1,152 @@
...
+} from "./createAwsKmsWallet";
+import {
+  createGcpKmsWallet,
+  type CreateGcpKmsWalletParams,
+} from "./createGcpKmsWallet";
+import { createLocalWallet } from "./createLocalWallet";
+import { getAwsKmsAccount } from "./getAwsKmsAccount";
+import { getGcpKmsAccount } from "./getGcpKmsAccount";
+
+const createSmartWallet = async ({
```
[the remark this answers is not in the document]
d4mr: Agreed this is confusing, I like your `generate` suggestion. It's still a little unclear because for AWS and GCP, generate would still have side effect (interactions with AWS and GCP).

src/db/wallets/createWalletDetails.ts
```diff
@@ -24,11 +29,39 @@ type CreateWalletDetailsParams = {
...
+    }
+  | {
+      type: "smart:gcp-kms";
+      gcpKmsResourcePath: string;
+      gcpApplicationCredentialPrivateKey: string; // will be encrypted and stored, pass plaintext to this function
+      gcpApplicationCredentialEmail: string;
+      accountSignerAddress: Address;
+
+      accountFactoryAddress?: Address;
+    }
```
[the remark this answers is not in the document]
d4mr: I think we shouldn't restrict to only a local wallet for the admin. IMO, I see smart wallets becoming the default wallet type for engine, so it should support everything we support for EOAs

We could make some modifications to the `create` endpoint and allow existing KMS wallets to become the signer. Underlying interface wouldn't need to change.

src/server/utils/wallets/legacyLocalCrypto.ts
```diff
@@ -0,0 +1,138 @@
...
+}
+
+export async function decryptJsonWallet({
+  password,
+  encryptedJson,
+}: { password: string; encryptedJson: string }): Promise<{
+  privateKey: string;
+  address: string;
+}> {
+  console.log(encryptedJson);
```
arcoraven, line 93: remove logging

src/server/utils/wallets/getLocalWallet.ts
```diff
@@ -16,6 +16,10 @@ interface GetLocalWalletParams {
   walletAddress: string;
 }
 
+/**
+ * @deprecated
+ * DEPRECATED: Use getLocalWalletAccount instead
+ */
```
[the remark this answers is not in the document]
d4mr: we can't because there's no way to expose the `privateKey` from the `Account` that we get from `getLocalWallet`. lmk if you think it's worth exposing it separately.

src/server/utils/wallets/createSmartWallet.ts
```diff
@@ -0,0 +1,152 @@
...
+} from "./createAwsKmsWallet";
+import {
+  createGcpKmsKey,
+  type CreateGcpKmsWalletParams,
+} from "./createGcpKmsWallet";
+import { createLocalWallet } from "./createLocalWallet";
+import { getAwsKmsAccount } from "./getAwsKmsAccount";
+import { getGcpKmsAccount } from "./getGcpKmsAccount";
+
+const getSmartWalletAddresses = async ({
```
arcoraven, line 19: It seems unnecessary for this function to accept `accountFactoryAddress` only to have the exact same value in the response. Can we simplify to just return the smartAccountAddress?

src/server/utils/wallets/createSmartWallet.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  const connectedAccount = await smartAccount.connect({
+    client: thirdwebClient,
+    personalAccount: adminAccount,
+  });
+
+  return {
+    accountAddress: connectedAccount.address,
+    accountFactoryAddress,
+  };
```
arcoraven, line 40: I think this whole function should be adjusted to return a connected account:
```typescript
getSmartAccount = async ({ adminAccount, accountFactoryAddress }): Promise<Account>
```

And have it return a connected smart account.
It's trivial to get the address (`(await getSmartAccount()).address`) and this function can be reused in `walletDetailsToAccount()` below where this code is repeated.

src/utils/account.ts
```diff
@@ -35,123 +40,183 @@ export const getAccount = async (args: {
...
+        sponsorGas: true,
+        factoryAddress: walletDetails.accountFactoryAddress ?? undefined,
+      });
+
+      const account = await unconnectedSmartWallet.connect({
+        client: thirdwebClient,
+        personalAccount: signerAccount,
+      });
+
+      return { account, adminAccount: signerAccount };
```
arcoraven, line 144: Nit: signerAccount and adminAccount should mean the same thing, so keep consistent and use one name everywhere. (No strong opinion which one)

I really want to avoid multiple names that mean the same thing since it's easy to get tripped up (we already have so many addresses to keep track of).

src/utils/account.ts
```diff
@@ -35,123 +40,183 @@ export const getAccount = async (args: {
...
+      });
+
+      return { account, adminAccount };
     }
     default:
       throw new Error(`Wallet type not supported: ${walletDetails.type}`);
   }
 };
 
+export const _adminAccountsCache = new Map<string, Account>();
```
arcoraven, line 169: Nit: Use LRUMap (see other usage on latest `main`)

src/utils/account.ts
```diff
@@ -35,123 +40,183 @@ export const getAccount = async (args: {
...
+  }
+
+  _adminAccountsCache.set(cacheKey, adminAccount);
+  return adminAccount;
+};
+
+const getAdminAccountCacheKey = (args: {
+  chainId: number;
+  accountAddress: Address;
+}) => `${args.chainId}-${args.accountAddress}`;
```
arcoraven, line 218: Shouldn't this be chain-agnostic if we store WalletDetails without chain? Though it doesn't hurt to include chain here too.

src/utils/cache/getWallet.ts
```diff
@@ -38,87 +38,50 @@ export const getWallet = async <TWallet extends EVMWallet>({
...
-      "BAD_REQUEST",
-    );
+  try {
+    walletDetails = await getWalletDetails({
+      pgtx,
+      address: walletAddress,
+    });
+  } catch (e) {
+    if (e instanceof WalletDetailsError) {
+      throw createCustomError(e.message, 400, "BAD_REQUEST");
```
arcoraven, line 50: Nit: Use StatusCodes everywhere for consistency (it'll be helpful if we're searching for where a certain status code is emitted). Or if you don't like em we should just switch off them entirely and use raw numbers.

src/utils/cache/getWallet.ts
```diff
@@ -127,6 +90,78 @@ export const getWallet = async <TWallet extends EVMWallet>({
...
 
+    case WalletType.smartAwsKms: {
+      if (accountAddress)
+        throw new Error(
+          "Smart backend wallet cannot be used to operate external smart account",
+        );
+
+      const splitArn = splitAwsKmsArn(walletDetails.awsKmsArn);
+
+      const personalWallet = new AwsKmsWallet({
```
arcoraven, line 101: Same nit: this is the adminWallet or signerWallet (pick one), so keep naming consistent.

src/worker/tasks/sendTransactionWorker.ts
```diff
@@ -95,9 +99,48 @@ const handler: Processor<any, void, string> = async (job: Job<string>) => {
...
+        resultTransaction = await _sendUserOp(job, transaction, adminAccount);
+      }
     } else {
-      resultTransaction = await _sendTransaction(job, transaction);
+      const account = await getAccount({
+        chainId: transaction.chainId,
+        from: transaction.from,
+      });
+
+      resultTransaction = await _sendTransaction(job, transaction, account);
```
arcoraven, line 143: It didn't seem like we needed to change this: getAccount can still be called inside _sendTransaction. If there's not a good reason to, let's revert this change.

src/worker/tasks/sendTransactionWorker.ts
```diff
@@ -95,9 +99,48 @@ const handler: Processor<any, void, string> = async (job: Job<string>) => {
...
+
+      if (!adminAccount) {
+        resultTransaction = {
+          ...transaction,
+          status: "errored",
+          errorMessage: "Failed to get admin account",
+        };
+      } else {
+        resultTransaction = await _sendUserOp(job, transaction, adminAccount);
+      }
```
arcoraven, line 136: Can all this just be moved inside _sendUserOp? Not sure if it needs to be out here, and it keeps all userOp logic inside the _sendUserOp function.

src/worker/tasks/sendTransactionWorker.ts
```diff
@@ -95,9 +99,48 @@ const handler: Processor<any, void, string> = async (job: Job<string>) => {
...
+        // do nothing, this might still be a regular backend wallet userop
+      }
+
+      try {
+        adminAccount = await getAccount({
+          chainId: transaction.chainId,
+          from: transaction.from,
+        });
+      } catch {
+        job.log(
```
arcoraven, line 123: Don't need this added logging. When this job errors, the `.on("failed")` listener at the bottom of this file already `job.log`'s all errors with the error message.

src/prisma/schema.prisma
```diff
@@ -94,6 +94,9 @@ model WalletDetails {
   gcpKmsResourcePath                 String? @map("gcpKmsResourcePath") @db.Text
   gcpApplicationCredentialEmail      String? @map("gcpApplicationCredentialEmail") /// if not available, default to: Configuration.gcpApplicationCredentialEmail
   gcpApplicationCredentialPrivateKey String? @map("gcpApplicationCredentialPrivateKey") /// if not available, default to: Configuration.gcpApplicationCredentialPrivateKey
+  // Smart Backend Wallet
+  accountSignerAddress               String? @map("accountSignerAddress") /// this, and either local, aws or gcp encryptedJson, are required for smart wallet
```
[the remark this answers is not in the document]
d4mr: the `WalletDetails` schema is very denormalized, so in the interest of maximum clarity, I think having an `account` prefix is better here.

Between `signer` and `admin`, I think `signer` is better. It leaves us with more flexibility later to allow importing SBW if we want.

src/db/wallets/createWalletDetails.ts
```diff
@@ -8,13 +9,17 @@ type CreateWalletDetailsParams = {
...
   | {
       type: "aws-kms";
       awsKmsKeyId?: string; // depcrecated and unused, todo: remove with next breaking change
       awsKmsArn: string;
 
-      awsKmsSecretAccessKey?: string; // will be encrypted and stored, pass plaintext to this function
-      awsKmsAccessKeyId?: string;
+      awsKmsSecretAccessKey: string; // will be encrypted and stored, pass plaintext to this function
+      awsKmsAccessKeyId: string;
```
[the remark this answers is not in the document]
d4mr: We decided to always denormalize and store credentials in the DB row, regardless of it being an override or coming from config. Users don't have to pass it in, the route fills in this from the config.

src/db/wallets/getWalletDetails.ts
```diff
@@ -1,20 +1,212 @@
...
+      },
+    });
+  } catch (e) {
+    if (e instanceof z.ZodError) {
+      throw new WalletDetailsError(
+        e.errors.map((error) => error.message).join(", "),
+      );
+    }
+    throw e;
+  }
```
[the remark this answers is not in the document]
d4mr: Agreed. Also unrelated, but I think we should move from a `util` mindset to a `lib` mindset. `util` has us thinking about one-off things that we can store to reuse later. `lib` would require us to be more intentional about categorizing and co-locating relevant functionality, so it's easier to discover/reuse later.

sdk/src/services/BackendWalletService.ts
```diff
@@ -665,6 +665,7 @@ export class BackendWalletService {
         requestBody: {
             message: string;
             isBytes?: boolean;
+            chainId?: number;
```
arcoraven, line 668: Note from our convo: can you check if signing on different chains result in the same output. If so, we could default to a testnet (Sepolia?).

src/server/utils/wallets/createLocalWallet.ts
```diff
@@ -13,7 +13,7 @@ interface CreateLocalWallet {
  * Create a local wallet with a random private key
  * Does not store the wallet in the database
  */
-export const createLocalWallet = async () => {
+export const createRandomLocalWallet = async () => {
```
arcoraven, line 16: Nit: "random" sounds unclear to me (especially since the other analogous KMS ones don't have it). createLocalWallet was fine, or generateLocalWallet?

src/lib/chain/chain-capabilities.ts
```diff
@@ -0,0 +1,60 @@
...
+] as const;
+
+export type Service = (typeof Services)[number];
+
+export type ChainCapabilities = Array<{
+  service: Service;
+  enabled: boolean;
+}>;
+
+// Create cache with 2048 entries and 5 minute TTL
```
arcoraven, line 21: Comment is wrong

src/lib/cache/swr.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    // Check if stale
+    const isStale = now > entry.validUntil;
+
+    // If stale, trigger background revalidation and return stale data
+    if (isStale) {
+      this.dedupedFetch(key, fetchFn).catch(() => {
+        // Silence background revalidation errors
+      });
```
arcoraven, line 39: So this catch makes it if any error occurs while fetching, the old data continues to be used.

Let's at least log a warning here. If this errors silently we won't know if the cache is stuck in a stale state.

arcoraven (approved):

PR #711, opened 2024-10-08
Simplify userop signing flow

joaquim-verges:
- use the SDK for creating and signing a userop instead of custom code

waiting on a SDK update, which will also add:

- respect custom gas set on the initial TX (overrides callGasLimit like we did before)
- Fix issue where the SDK would wait for the account to be fully deployed before creating another userop for that same account (disabling that behavior for engine, it doesn't make sense in a server context)

---

## PR-Codex overview
This PR focuses on updating the `thirdweb` dependency, enhancing the user operation handling in the `sendTransactionWorker.ts`, and removing an obsolete test file.

### Detailed summary
- Deleted the test file: `src/utils/transaction/userOperation.tstest/e2e/tests/write.test.ts`
- Updated `thirdweb` version from `^5.60.1` to `5.61.2` in `package.json`.
- Added new imports from `thirdweb`: `getContract`, `readContract`, `createAndSignUserOp`.
- Modified the structure of `queuedTransaction` to include additional properties: `from`, `userProvidedAccountFactoryAddress`, `overrides`.
- Enhanced error handling for fetching the account factory address.
- Updated user operation signing logic to include `accountFactoryAddress` and `overrides`.
- Removed the obsolete `generateSignedUserOperation` function call.


src/utils/transaction/userOperation.ts
```diff
@@ -1,134 +0,0 @@
...
-  const unsignedOp = (await createUnsignedUserOp({
-    transaction: userOpCall,
-    accountContract: smartAccountContract,
-    sponsorGas: true,
-    factoryContract: accountFactoryContract,
-    adminAddress: signerAddress,
-  })) as UserOperation; // TODO support entrypoint v0.7 accounts
-
-  // Handle gas limit override.
-  if (overrides?.gas) {
```
arcoraven, line 97: As per @0xFirekeeper, this is being used by one of our large customers. We may still need gas override options in userOps.

joaquim-verges: Handled in the SDK now.

It will respect whatever gas you set in the initial tx you're trying to bundle

arcoraven (approved):

PR #715, opened 2024-10-09
Expose `x-account-salt` header to handle multiple smart wallets for single admin wallet

joaquim-verges:
to test: 

```bash
cd test/e2e
bun test userop --timeout 12000
```

---

## PR-Codex overview
This PR focuses on adding support for `accountSalt` in various transaction-related functionalities, enhancing the ability to predict smart account addresses, and updating the `thirdweb` dependency.

### Detailed summary
- Updated `thirdweb` version from `5.61.2` to `5.61.3`.
- Added `accountSalt` property in several TypeScript files and schemas.
- Modified functions to handle `accountSalt` in transaction processing.
- Updated tests to include scenarios involving `accountSalt`.
- Adjusted import statements for consistency.


src/server/routes/contract/extensions/accountFactory/write/createAccount.ts
```diff
@@ -63,37 +69,53 @@ export const createAccount = async (fastify: FastifyInstance) => {
...
-        simulateTx,
-        extension: "account-factory",
-        deployedContractAddress: deployedAddress,
-        deployedContractType: "account",
-        idempotencyKey,
+      // if extraData is not a hex string, convert it to a hex string
+      // this is the same transformation that is done in the SDK
+      // for predictAddress and createAndSignUserOp
+      // but needed here because we're calling the raw autogenerated abi function
+      const accountSaltHex =
```
arcoraven, line 95: I assume you want this "hexify if not hex" logic above the `predictAddress` code above.

src/server/routes/contract/extensions/accountFactory/write/createAccount.ts
```diff
@@ -63,37 +69,53 @@ export const createAccount = async (fastify: FastifyInstance) => {
...
+        extraData && isHex(extraData)
+          ? extraData
+          : stringToHex(extraData ?? "");
+
```
arcoraven, line 99: I see `data` is required in `factoryCreateAccount` so you're hexifying `""` => `0x` if extraData is not provided.

src/server/routes/contract/extensions/accountFactory/read/predictAccountAddress.ts
```diff
@@ -51,16 +53,16 @@ export const predictAccountAddress = async (fastify: FastifyInstance) => {
...
-      const accountAddress =
-        await contract.accountFactory.predictAccountAddress(
-          adminAddress,
-          extraData,
-        );
+
+      const accountAddress = await predictAddress({
+        factoryContract,
+        adminAddress,
+        accountSalt: extraData,
```
arcoraven, line 64: Should we also "hexify if not hex" here?

src/server/routes/contract/extensions/accountFactory/write/createAccount.ts
arcoraven: Also if we expect to need this logic in all methods, I wonder if we should write a `parseAccountAbstractionHeaders()` function that returns the AA headers with "hexified" accountSalt.

src/server/schemas/transaction/index.ts
```diff
@@ -178,6 +178,8 @@ export const TransactionSchema = Type.Object({
   ]),
   signerAddress: Type.Union([AddressSchema, Type.Null()]),
   accountAddress: Type.Union([AddressSchema, Type.Null()]),
+  accountSalt: Type.Union([Type.String(), Type.Null()]),
+  accountFactoryAddress: Type.Union([AddressSchema, Type.Null()]),
```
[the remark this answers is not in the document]
arcoraven: This looks right if you're looking to add nullable fields here.

src/server/schemas/wallet/index.ts
```diff
@@ -27,6 +27,12 @@ export const walletWithAAHeaderSchema = Type.Object({
     description:
       "Smart account factory address. If omitted, engine will try to resolve it from the chain.",
   }),
+  "x-account-salt": Type.Optional(
+    Type.String({
+      description:
+        "Smart account salt in hex format. Useful when creating multiple accounts with the same admin and only needed when deploying the account as part of a userop.",
```
arcoraven, line 33: Hex format is required unless we hexify non-hex strings. If hex is required, I'd recommend validating it at the API level:
```typescript
Type.RegExp(/^0x[a-fA-F0-9]*$/, {...})
```
instead of
```typescript
Type.String(...)
```

src/utils/transaction/types.ts
```diff
@@ -41,6 +41,7 @@ export type InsertedTransaction = {
   // User Operation
   signerAddress?: Address;
   accountAddress?: Address;
+  accountSalt?: string;
```
arcoraven, line 44: Maybe use `Hex` here? You may need to type it `as Hex` when inserting it but it's prob more correct.

src/server/routes/contract/extensions/accountFactory/write/createAccount.ts
joaquim-verges: its actually only in this case that we need this, everywhere else i do it for you inside the SDK. But here we're calling the 'raw' abi method for createAccount which does not do any pre processing, so i have to do it before calling that raw abi method.

considering making an extension just to do this lil part for you internally

src/server/routes/contract/extensions/accountFactory/read/predictAccountAddress.ts
joaquim-verges: see above, its done internally

src/server/schemas/wallet/index.ts
joaquim-verges: let me update the comment actually, it does not need to be in hex format anymore

src/utils/transaction/types.ts
joaquim-verges: same as above, I actually want ppl to pass us strings without having to worry about hexing.

updated the header description

src/server/routes/contract/extensions/accountFactory/write/createAccount.ts
arcoraven: I see, but it would still work to put this higher? It just isn't obvious to someone who doesn't know the SDK behavior that this replicates logic already done in the handler above. 

Alternatively, can createAccount do the same hexify transformation so this logic can be removed?

src/utils/transaction/types.ts
arcoraven: I see, it may not be hexified yet when it gets stored here.

joaquim-verges:
### Merge activity

* **Oct 11, 3:58 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thirdweb-dev/engine/715).

PR #737, opened 2024-10-28
fix: Missing low wallet balance webhook

arcoraven:
## PR-Codex overview
This PR focuses on refactoring the webhook handling for wallet balance notifications. It introduces a new type `BackendWalletBalanceWebhookParams`, modifies existing webhook event processing, and implements a mechanism to notify when the wallet balance is low.

### Detailed summary
- Renamed `WalletBalanceWebhookSchema` to `BackendWalletBalanceWebhookParams`.
- Updated `sendWebhookWorker.ts` to handle `WebhooksEventTypes.BACKEND_WALLET_BALANCE`.
- Added `_enqueueBackendWalletBalanceWebhook` method in `SendWebhookQueue`.
- Implemented `_notifyIfLowBalance` in `mineTransactionWorker` to check and notify low wallet balances.
- Removed the old `sendBalanceWebhook` function in `webhook.ts`.


src/utils/webhook.ts
```diff
@@ -69,67 +62,11 @@ export const sendWebhookRequest = async (
...
     return {
       ok: false,
       status: 500,
-      body: e.toString(),
+      body: prettifyError(e),
     };
   }
 };
-
-// TODO: Add retry logic upto
```
arcoraven, line 81: This was legacy from v1 and unused. Migrated to v2 queues and trigger it from `mineTransactionWorker`

src/worker/tasks/mineTransactionWorker.ts
```diff
@@ -77,12 +83,6 @@ const handler: Processor<any, void, string> = async (job: Job<string>) => {
         walletAddress: getAddress(resultTransaction.from),
       },
     });
-    logger({
```
arcoraven, line 80: Removing unneeded logging.

src/worker/tasks/mineTransactionWorker.ts
```diff
@@ -257,6 +264,66 @@ const _mineUserOp = async (
   };
 };
 
+const _notifyIfLowBalance = async (transaction: MinedTransaction) => {
+  const { isUserOp, chainId, from } = transaction;
+  if (isUserOp) {
+    // Skip for userOps since they may not use the wallet's gas balance.
```
edwardysun, line 270: Doesn't infinigods use smart wallet transactions?

edwardysun (approved):

src/worker/tasks/mineTransactionWorker.ts
arcoraven: UserOps are primarily gas sponsored via Engine.
This check is useful for users sending non userOp transactions from backend wallets, which they're using too.

dcapitator:
The wallet balance webhook is built to serve a case when we have few wallets . 

What if we have thousands and we only want to listen to funding wallets or only wallets used for gasless Txs ??

Can that feature be considered ? Instead of listening to all backend wallets then we have an array of wallets.

PR #762, opened 2024-11-06
refactor all int chainId to string

d4mr:
## PR-Codex overview
This PR focuses on changing the data type of `chainId` from `Int` to `String` across various models and functions, ensuring consistent handling of `chainId` as a string throughout the codebase.

### Detailed summary
- Updated `chainId` to `String` in `ContractEventLogs`, `ContractTransactionReceipts`, and `ChainIndexers` models.
- Changed `chainId` handling to `toString()` in multiple queries and functions.
- Parsed `chainId` as `Number.parseInt()` where necessary.
- Modified SQL migration to alter `chainId` data type to `TEXT`.

### DIFF
--- a/src/db/chainIndexers/getChainIndexer.ts
+++ b/src/db/chainIndexers/getChainIndexer.ts
@@ -43,7 +43,7 @@ export const getBlockForIndexing = async ({
     FROM
       "chain_indexers"
     WHERE
-      "chainId"=${Prisma.sql`${chainId}`}
+      "chainId"=${Prisma.sql`${chainId.toString()}`}
     FOR UPDATE NOWAIT
   `;
   return lastIndexedBlock[0].lastIndexedBlock;
--- a/src/db/chainIndexers/getChainIndexer.ts
+++ b/src/db/chainIndexers/getChainIndexer.ts
@@ -1,3 +1,4 @@
+import { Prisma } from "@prisma/client";
 import type { PrismaTransaction } from "../../schema/prisma";
 import { getPrismaWithPostgresTx } from "../client";
 
@@ -14,7 +15,7 @@ export const getLastIndexedBlock = async ({
 
   const indexedChain = await prisma.chainIndexers.findUnique({
     where: {
-      chainId,
+      chainId: chainId.toString(),
     },
   });
 
@@ -42,7 +43,7 @@ export const getBlockForIndexing = async ({
     FROM
       "chain_indexers"
     WHERE
-      "chainId"=${chainId}
+      "chainId"=${Prisma.sql`${chainId}`}
     FOR UPDATE NOWAIT
   `;
   return lastIndexedBlock[0].lastIndexedBlock;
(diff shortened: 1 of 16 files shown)


src/db/contractEventLogs/getContractEventLogs.ts
```diff
@@ -148,22 +150,22 @@ export const getEventLogsByCursor = async ({
                 { createdAt: { gt: cursorObj.createdAt } },
                 {
                   createdAt: { equals: cursorObj.createdAt },
-                  chainId: { gt: cursorObj.chainId },
```
d4mr, line 151: this feels accidental, we shouldn't need to compare `gt` for chainId right?

src/db/contractTransactionReceipts/getContractTransactionReceipts.ts
```diff
@@ -118,16 +120,16 @@ export const getTransactionReceiptsByCursor = async ({
                 { createdAt: { gt: cursorObj.createdAt } },
                 {
                   createdAt: { equals: cursorObj.createdAt },
-                  chainId: { gt: cursorObj.chainId },
```
d4mr, line 121: here too, feels copy+pasted

src/db/chainIndexers/getChainIndexer.ts
```diff
@@ -42,7 +43,7 @@ export const getBlockForIndexing = async ({
     FROM
       "chain_indexers"
     WHERE
-      "chainId"=${chainId}
```
d4mr, line 45: needed to wrap `${chainId}` in quotes to work for a SQL string comparison. Opted to use `Prisma.sql` which is the recommended way of templating in SQL string.

arcoraven: Hm why can't this line just be:
```
"chainId"="${chainId.toString()}"
```

src/db/contractEventLogs/getContractEventLogs.ts
arcoraven: Actually, leave this as "gt" (even though the behavior is different with strings). It's a safer change than updating this to equals. If using eq here it matches many more rows than it would before.

src/db/chainIndexers/getChainIndexer.ts
d4mr: yeah that works, and given that way know `chainId` is sanitised it should be fine, but just a standard best practice to not use raw string interpolation with SQL

arcoraven: Lets simplify it. From how I'm reading Prisma docs ([source](https://www.prisma.io/docs/orm/prisma-client/using-raw-sql/raw-queries#queryraw)):
> The method is implemented as a [tagged template](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates), which allows you to pass a template literal where you can easily insert your [variables](https://www.prisma.io/docs/orm/prisma-client/using-raw-sql/raw-queries#using-variables). In turn, Prisma Client creates prepared statements that are safe from SQL injections:
```
const email = 'emelie@prisma.io'
const result = await prisma.$queryRaw`SELECT * FROM User WHERE email = ${email}`
```

arcoraven (approved):

PR #766, opened 2024-11-08
fix: retry-failed creates undefined nonce value

arcoraven:
## PR-Codex overview
This PR focuses on improving error handling in the transaction processing logic and refining the retry mechanism for failed transactions.

### Detailed summary
- Updated error handling in `withErrorHandler` to support string errors.
- Simplified the error response structure.
- Enhanced the transaction retry logic in `sendTransactionWorker` by omitting unnecessary fields.
- Improved error message handling for errored transactions.

### DIFF
--- a/src/server/middleware/error.ts
+++ b/src/server/middleware/error.ts
@@ -53,7 +53,17 @@ const isZodError = (err: unknown): boolean => {
 
 export const withErrorHandler = async (server: FastifyInstance) => {
   server.setErrorHandler(
-    (error: Error | CustomError | ZodError, request, reply) => {
+    (error: string | Error | CustomError | ZodError, request, reply) => {
+      if (typeof error === "string") {
+        return reply.status(StatusCodes.INTERNAL_SERVER_ERROR).send({
+          error: {
+            statusCode: 500,
+            code: "INTERNAL_SERVER_ERROR",
+            message: error || ReasonPhrases.INTERNAL_SERVER_ERROR,
+          },
+        });
+      }
+
       // Ethers Error Codes
       if (parseEthersError(error)) {
         return reply.status(StatusCodes.BAD_REQUEST).send({
@@ -103,7 +113,7 @@ export const withErrorHandler = async (server: FastifyInstance) => {
           StatusCodes.INTERNAL_SERVER_ERROR;
 
         const message = error.message ?? ReasonPhrases.INTERNAL_SERVER_ERROR;
-        reply.status(statusCode).send({
+        return reply.status(statusCode).send({
           error: {
             code,
             message,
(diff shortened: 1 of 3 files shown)


src/server/middleware/error.ts
```diff
@@ -53,7 +53,17 @@ const isZodError = (err: unknown): boolean => {
 
 export const withErrorHandler = async (server: FastifyInstance) => {
   server.setErrorHandler(
-    (error: Error | CustomError | ZodError, request, reply) => {
+    (error: string | Error | CustomError | ZodError, request, reply) => {
+      if (typeof error === "string") {
```
arcoraven, line 57: Currently a thrown string is not logged.
While we shouldn't throw raw strings, this update at least prints the string value.

src/worker/tasks/sendTransactionWorker.ts
```diff
@@ -78,17 +78,15 @@ const handler: Processor<string, void, string> = async (job: Job<string>) => {
   // For example, the developer retried all failed transactions during an RPC outage.
   // An errored queued transaction (resendCount = 0) is safe to retry: the transaction wasn't sent to RPC.
   if (transaction.status === "errored" && resendCount === 0) {
+    const { errorMessage, ...omitted } = transaction;
     transaction = {
-      ...{
-        ...transaction,
-        nonce: undefined,
```
arcoraven, line 84: This was the bug. By setting values to undefined, they were _explicitly_ set to undefined in the type. This is unexpected because the typing implies undefined is not a valid value, which leads to weird edge cases like the code not handling undefined values.

arcoraven: The fix here was to flatten this code to override certain fields only.

Note: 
- nonce doesn't need to be unset since we're calling `_sendTransaction()` again which generates a new nonce as expected.
- gas fees don't need to be unset since overrides were moved to a separate nested object.

src/worker/tasks/sendTransactionWorker.ts
```diff
@@ -246,11 +244,14 @@ const _sendUserOp = async (
...
+    const errorMessage = wrapError(error, "Bundler").message;
+    const erroredTransaction: ErroredTransaction = {
       ...queuedTransaction,
       status: "errored",
-      errorMessage: wrapError(error, "Bundler").message,
-    } satisfies ErroredTransaction;
+      errorMessage,
+    };
+    job.log(`Failed to populate transaction: ${errorMessage}`);
+    return erroredTransaction;
```
arcoraven, line 254: Unrelated fix, I previously removed the job.log() here but now I want it back. It's important for us to know if an error was thrown during simulation/estimation or when sending the tx.

PR #773, opened 2024-11-15
feat: batching for reads and smart-account writes

d4mr:
Test `read` with example curl:

```
curl --request POST \
  --url http://localhost:3005/contract/421614/read-batch \
  --header 'Authorization: Bearer {{authToken}}' \
  --header 'content-type: application/json' \
  --data '{
  "calls": [
    {
      "contractAddress": "0x75faf114eafb1BDbe2F0316DF893fd58CE46AA4d",
      "functionName": "name"
    },
    {
      "contractAddress": "0x75faf114eafb1BDbe2F0316DF893fd58CE46AA4d",
      "functionName": "totalSupply"
    }
  ]
}'
```

Test `write` cURL example:
```
curl --request POST \
  --url http://localhost:3005/backend-wallet/84532/send-transactions-atomic \
  --header 'Authorization: Bearer {{authToken}}' \
  --header 'Content-Type: application/json' \
  --header 'x-backend-wallet-address: <use a smart backend wallet> \
  --data '{
  "transactions": [
    {
      "toAddress": "0x5175E053bE7BDC74896C0D8675590FfbCbfe2E73",
      "value": "0",
      "data": "0xe5071b8e"
    },
    {
      "toAddress": "0x5175E053bE7BDC74896C0D8675590FfbCbfe2E73",
      "value": "0",
      "data": "0xe5071b8e"
    }
  ]
}
```
`0xe5071b8e` is the function selector for `incrementCount()`, and `0x5175E053bE7BDC74896C0D8675590FfbCbfe2E73` is deployed on [base sepolia](https://sepolia.basescan.org/address/0x5175E053bE7BDC74896C0D8675590FfbCbfe2E73). On checking status of queued transaction, and viewing event logs for the sent transaction, logs from `0x51..73` should be visilble.

image

---

## PR-Codex overview
This PR introduces support for batch operations in transaction processing, enhancing the ability to handle multiple transactions atomically. It includes new schemas, routes, and methods for batch reading and sending transactions, as well as updates to existing transaction handling.

### Detailed summary
- Added `batchOperations` field to various schemas and types.
- Introduced `sendTransactionBatchAtomic` route for atomic transaction processing.
- Created `readBatchRoute` for batch reading from multiple contracts.
- Updated transaction handling to support batch operations.
- Enhanced error handling for wallet details retrieval.
- Modified OpenAPI URL in `generate-sdk.ts` for local development.
- Implemented caching for wallet details to improve performance.


d4mr:
changelog

src/server/routes/contract/read/read-batch.ts
```diff
@@ -0,0 +1,170 @@
...
+      const { calls, multicallAddress = MULTICALL3_ADDRESS } = request.body;
+
+      const chainId = await getChainIdFromChain(chainSlug);
+      const chain = await getChain(chainId);
+
+      try {
+        // Encode each read call
+        const encodedCalls = await Promise.all(
+          calls.map(async (call) => {
+            const contract = await getContract({
```
kien-ngo, line 86: You don't have to await getContract btw

src/server/routes/contract/read/read-batch.ts
```diff
@@ -0,0 +1,170 @@
...
+  chain: Type.String(),
+});
+
+type RouteGeneric = {
+  Params: { chain: string };
+  Body: Static<typeof readMulticallRequestSchema>;
+  Reply: Static<typeof responseSchema>;
+};
+
+export async function readMulticall(fastify: FastifyInstance) {
```
arcoraven, line 58: Small nit and non-blocking: I started calling route names `readMulticallRoute`. The reason is we have so many exported functions that end up having the same name (for example: `createAccessToken` might be the route, or the cached call, or the raw db call, who knows?)

So having that suffix is nice for clarity.

src/server/routes/contract/read/read-batch.ts
```diff
@@ -0,0 +1,170 @@
...
+const readCallRequestItemSchema = Type.Object({
+  contractAddress: Type.String(),
+  functionName: Type.String(),
+  functionAbi: Type.Optional(Type.String()),
+  args: Type.Optional(Type.Array(Type.Any())),
+});
+
+const readMulticallRequestSchema = Type.Object({
+  calls: Type.Array(readCallRequestItemSchema),
+  multicallAddress: Type.Optional(Type.String()),
```
arcoraven, line 36: Why would someone want to override this?

src/server/routes/contract/read/read-batch.ts
```diff
@@ -0,0 +1,170 @@
...
+};
+
+export async function readMulticall(fastify: FastifyInstance) {
+  fastify.route<RouteGeneric>({
+    method: "POST",
+    url: "/contract/:chain/read-batch",
+    schema: {
+      summary: "Batch read from multiple contracts",
+      description:
+        "Execute multiple contract read operations in a single call using Multicall3",
```
arcoraven, line 65: Lets sync on how we expose the "multicall" concept. I think we either make it clear this is a multicall (e.g. the route name is multicall) or remove it entirely. It's confusing to have both "read batch" and "multicall" used interchangeably.

src/server/routes/contract/read/read-batch.ts
```diff
@@ -0,0 +1,170 @@
...
+              contract,
+              method,
+              params: call.args || [],
+              // stubbing gas values so that the call can be encoded
+              maxFeePerGas: 30n,
+              maxPriorityFeePerGas: 1n,
+              value: 0n,
+            });
+
+            const calldata = await resolvePromisedValue(transaction.data);
```
arcoraven, line 106: Don't use resolvePromisedValue. I should add a `@deprecated` tag on the function. Use this instead:
```suggestion
const callData = await encode(transaction);
```

src/server/routes/contract/read/read-batch.ts
```diff
@@ -0,0 +1,170 @@
...
+              (await resolveMethod(call.functionName)(contract));
+
+            const transaction = prepareContractCall({
+              contract,
+              method,
+              params: call.args || [],
+              // stubbing gas values so that the call can be encoded
+              maxFeePerGas: 30n,
+              maxPriorityFeePerGas: 1n,
+              value: 0n,
```
arcoraven, line 103: Do these need to be stubbed? What happens if we leave them undefined?

src/server/routes/contract/read/read-batch.ts
```diff
@@ -0,0 +1,170 @@
...
+          );
+
+          const decoded = decodeAbiParameters(
+            outputs,
+            returnData as `0x${string}`,
+          );
+
+          return {
+            success,
+            result: success ? bigNumberReplacer(decoded) : null,
```
arcoraven, line 152: We should not need bigNumberReplacer. We're on v5 SDK which does not rely on ethers.

src/server/routes/contract/read/read-batch.ts
```diff
@@ -0,0 +1,170 @@
...
+export async function readMulticall(fastify: FastifyInstance) {
+  fastify.route<RouteGeneric>({
+    method: "POST",
+    url: "/contract/:chain/read-batch",
+    schema: {
+      summary: "Batch read from multiple contracts",
+      description:
+        "Execute multiple contract read operations in a single call using Multicall3",
+      tags: ["Contract"],
+      operationId: "readMulticall",
```
arcoraven, line 67: `readBatch` is preferred since "multicall" is the underlying approach but the use case is a batch read.

arcoraven (approved):

joaquim-verges (approved):

src/server/routes/backend-wallet/send-transactions-atomic.ts
```diff
@@ -0,0 +1,151 @@
...
+
+export async function sendTransactionsAtomicRoute(fastify: FastifyInstance) {
+  fastify.route<{
+    Params: Static<typeof walletChainParamSchema>;
+    Body: Static<typeof requestBodySchema>;
+    Reply: Static<typeof transactionWritesResponseSchema>;
+    Querystring: Static<typeof requestQuerystringSchema>;
+  }>({
+    method: "POST",
+    url: "/backend-wallet/:chain/send-transactions-atomic",
```
arcoraven, line 48: There's already a `/send-transactions-batch`. IMO introducing `/send-transactions-atomic` will confuse the dev.

WDYT about reusing the same handler and branching the logic if SBW? In the docs we'd then need to clarify "Order is guaranteed for SBW only. If non-SBW, order is not guaranteed".

If you still want two separate paths, I'd prefer avoiding a new term (atomic) and using simpler language, like `/send-transaction-batch-ordered`. Or `/send-userop-batch` to make it clear it's for smart wallets.

joaquim-verges, line 48: maybe send-batch-transaction ? we kinda use that language already in the sdk and docs

src/server/routes/backend-wallet/send-transactions-atomic.ts
```diff
@@ -0,0 +1,151 @@
...
+  transactions: Type.Array(
+    Type.Object({
+      toAddress: Type.Optional(AddressSchema),
+      data: Type.String({
+        examples: ["0x..."],
+      }),
+      value: Type.String({
+        examples: ["10000000"],
+      }),
+    }),
```
arcoraven, line 36: Non-blocking: These are the minimal args of a "raw transaction" and may be useful to DRY since it's used in a few places. Then we can add descriptions too.

src/server/routes/backend-wallet/send-transactions-atomic.ts
```diff
@@ -0,0 +1,151 @@
...
+  type ParsedWalletDetails,
+  WalletDetailsError,
+} from "../../../shared/db/wallets/get-wallet-details";
+import { createCustomError } from "../../middleware/error";
+
+const requestBodySchema = Type.Object({
+  transactions: Type.Array(
+    Type.Object({
+      toAddress: Type.Optional(AddressSchema),
+      data: Type.String({
```
arcoraven, line 30: Use `HexSchema`

src/server/routes/backend-wallet/send-transactions-atomic.ts
```diff
@@ -0,0 +1,151 @@
...
+        }
+      }
+
+      if (transactionRequests.length === 0) {
+        throw createCustomError(
+          "No transactions provided",
+          StatusCodes.BAD_REQUEST,
+          "NO_TRANSACTIONS_PROVIDED",
+        );
+      }
```
arcoraven, line 120: You should be able to assert this in TypeBox: `{ minItems: 1 }`

src/server/routes/backend-wallet/send-transactions-atomic.ts
```diff
@@ -0,0 +1,151 @@
...
+import { createCustomError } from "../../middleware/error";
+
+const requestBodySchema = Type.Object({
+  transactions: Type.Array(
+    Type.Object({
+      toAddress: Type.Optional(AddressSchema),
+      data: Type.String({
+        examples: ["0x..."],
+      }),
+      value: Type.String({
```
arcoraven, line 33: Use `WeiAmountStringSchema`

src/server/routes/backend-wallet/send-transactions-atomic.ts
d4mr: We already have a send-transaction-batch endpoint which just queues up all the transactions and sends them non atomically

src/shared/db/wallets/get-wallet-details.ts
```diff
@@ -1,3 +1,4 @@
+import LruMap from "mnemonist/lru-map";
```
arcoraven, line 1: Nit: LRUMap to be consistent with other places we use this

src/server/routes/contract/read/read-batch.ts
```diff
@@ -0,0 +1,170 @@
...
+          chain,
+          address: multicallAddress,
+          client: thirdwebClient,
+        });
+
+        // Execute batch read
+        const results = await readContract({
+          contract: multicall,
+          method: MULTICALL3_AGGREGATE_ABI,
+          params: [encodedCalls],
```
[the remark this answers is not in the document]
d4mr: There's definitely a limit, but failing and forwarding the error should be fine, no?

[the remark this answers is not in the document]
joaquim-verges: Yeah should be fine

src/server/routes/contract/read/read-batch.ts
```diff
@@ -0,0 +1,170 @@
...
+  chain: Type.String(),
+});
+
+type RouteGeneric = {
+  Params: { chain: string };
+  Body: Static<typeof readMulticallRequestSchema>;
+  Reply: Static<typeof responseSchema>;
+};
+
+export async function readMulticallRoute(fastify: FastifyInstance) {
```
arcoraven, line 62: Nit: I consider multicall the implementation, and any interface higher than this method shouldn't need to know about it. So var/fn names (even if not user facing) don't need to have multicall in them. Maybe `batchReadRoute` or `readBatchRoute`?

src/server/routes/backend-wallet/send-transactions-atomic.ts
joaquim-verges: Why don't you edit that endpoint then? Just do an actual batch if it's a smart backend wallet

src/worker/tasks/send-transaction-worker.ts
```diff
@@ -143,7 +143,14 @@ const _sendUserOp = async (
...
   const toAddress = to ?? target;
-  assert(toAddress, "Invalid transaction parameters: to");
+
+  if (queuedTransaction.batchOperations) {
+    queuedTransaction.batchOperations.map((op) => {
+      assert(op.to, "Invalid transaction parameters: to");
+    });
+  } else {
+    assert(toAddress, "Invalid transaction parameters: to");
+  }
```
arcoraven, line 153: For my understanding:
- in a single userOp, `toAddress` is set.
- in a batch userOp, `toAddress` is not set but it must be set on every batch operation.

What about contract deploys? Is there a toAddress on a userOp to deploy a contract?

src/worker/tasks/send-transaction-worker.ts
```diff
@@ -199,17 +206,25 @@ const _sendUserOp = async (
...
+        }))
+      : [
+          {
+            client: thirdwebClient,
+            chain,
+            ...queuedTransaction,
+            ...overrides,
+            to: getChecksumAddress(toAddress),
+          },
+        ];
```
arcoraven, line 223: We're explicitly ignoring `overrides` in the batch. That said, I recall discussion that gas/gasFee overrides for userOps don't make sense and we should remove them anyway. We should pick one or the other: this code does it both ways (overrides for single userOp, no overrides for batch userOps).

arcoraven: _Non-blocking comment (maybe cleanup for later)_

Single vs batch userOp flows should overlap as much as possible: there should be minimal difference other than the array of transactions being submitted.

That said I think all the "single vs batch userOp" logic should be in one place.

```
// pseudocode
let transactions = []
if (batchOperations) {
  for (op of batchOperations) {
    assert(op.to)
    transactions.push({ ...op, chain, client })
  }
} else {
  transactions.push({ to, data, value, chain, client })
}

// ...the rest of the code doesn't care about batchOperations
```

src/server/routes/backend-wallet/send-transactions-atomic.ts
d4mr: I considered modifying the `send-transaction-batch` endpoint, but decided against it, because:

1. There's legitimate use-cases for non-atomic batching, even when using smart backend wallets. Batching atomically you might run into block gas limit, no limits on non-atomic batching though

2. Because of above reason, we shouldn't auto-force atomic batch on SBW. This might break user expectations, so instead having a `batch?: boolean` option would be good. This isn't easy to add, because existing endpoint takes an expects an array as the transaction body, no easy way to add to request body.
(We can still add to queryParam)

3. Another argument against automatic atomic batching when SBW: we also want to support atomic batching for regular EOA + Account Address style smart accounts.

4. There's also a couple of things don't make sense to share between the two batching strategies. Eg, `idempotency-key`. Non-atomic batching returns multiple queue-ids (`queueIds: string[]`), so an idempotency key doesn't work there, but it does make sense to include when using atomic batching, since there's only one queue-id returned.

5. `shouldSimulate` is not available as a query-param for non-atomic batch, due to latency of individually simulating transactions. But we can allow this for atomic batch.

6. As mentioned before: non atomic batch returns an array of `queueIds`, and atomic batch only returns a single queueId. If we automatically move to atomic batching on the same endpoint, and start returning only a single `queueId` in the array, client code that expects `transactions.length == queueIds.length` might break

Given the number of special cases to handle if we were to merge endpoints, I think having them separate is best.

d4mr: answered why not the same handler in a question above, Joaquim had a similar question.

I do agree the name for the other path could be improved, but I ran out of ideas.

`send-userop-batch` could be misleading since we're creating a single UserOp rather than batching multiple UserOps.

I do like keeping "atomic" in the endpoint name since it's a widely understood term beyond web3 - developers regularly encounter it in databases, filesystems, and other contexts where operations need to succeed or fail as one unit, and it's more precise than "ordered".

I was initially thinking `/send-transaction-batch-atomic` to go along with the existing `/send-transaction-batch`, but felt it would be too long.

src/worker/tasks/send-transaction-worker.ts
d4mr: for contract deployment, it should be `0x0000000000000000000000000000000000000000`

src/worker/tasks/send-transaction-worker.ts
d4mr: > gas/gasFee overrides for userOps don't make sense

the current overrides don't work for userOps correct, but overrides in general can still be changed to use with userOps:

```
type UserOpGasOverrides = {
  callGasLimit?: bigint
  verificationGasLimit?: bigint
  preVerificationGas?: bigint
  maxFeePerGas?: bigint  
  maxPriorityFeePerGas?: bigint
}
```

even in the batch userop flow, there's only a single UserOp being constructed. And gas overrides are applied per UserOp, not per transaction inside the userop.

src/worker/tasks/send-transaction-worker.ts
arcoraven: I thought contract deployment `to = null`? Something to test

src/server/routes/backend-wallet/send-transactions-atomic.ts
arcoraven: You listed enough differences to justify a different endpoint rather than overloading one. :+1:

More thoughts on names:
- /send-ordered-transactions
- /send-userop-batch
- /send-transaction-sequence
I don't have a strong opinion against atomic, seems clear too.

joaquim-verges: ugh i see. is anyone actually using the non-atomic batch endpoint though? seems very niche when you can just send 2 api calls

arcoraven (approved):

PR #784, opened 2024-11-29
feat: add support for gasPrice in transaction overrides and schemas

d4mr:


src/server/schemas/txOverrides.ts
```diff
@@ -10,8 +10,11 @@ export const txOverridesSchema = Type.Object({
         description: "Gas limit for the transaction",
       }),
 
-      // Overriding `gasPrice` is currently not supported.
-
+      gasPrice: Type.Optional({
```
d4mr, line 13: we could also allow for a boolean `true` value here, which if detected, engine calculates the right gas amount to send

PR #819, opened 2024-12-16
Address balance listener worker

nischitpra:
## PR-Codex overview
This PR introduces a new `config` column in the `webhooks` and `configuration` tables, enhances the webhook creation process to include configuration details, and implements an `addressBalanceListener` that triggers webhooks based on address balance thresholds.

### Detailed summary
- Added `config` column to `webhooks` and `configuration` tables.
- Updated `get-configuration.ts` to initialize `addressBalanceListenerCronSchedule`.
- Modified `toWebhookSchema` to include `config`.
- Enhanced `createWebhookRoute` to handle `config`.
- Implemented `addressBalanceListener` to track address balance changes.
- Developed `trackAddressBalance` to enqueue webhooks based on balance thresholds.
- Created e2e tests for the address balance listener functionality.


src/shared/db/configuration/get-configuration.ts
```diff
@@ -236,6 +236,10 @@ export const getConfiguration = async (): Promise<ParsedConfig> => {
     config = await updateConfiguration({
       indexerListenerCronSchedule: "*/5 * * * * *",
     });
+  } else if (!config.addressBalanceListenerCronSchedule) {
+    config = await updateConfiguration({
+      addressBalanceListenerCronSchedule: "0 */5 * * * *",
```
nischitpra, line 241: Is 5 minutes a good poll interval?

d4mr (changes requested):
In the current implementation, in order to create a balance listener, a user creates a webhook with a config (newly added field), where a user can specify options like `address`, `chainId` and `threshold`).

The webhooks table is mostly intended to store URLs that engine should call, and business logic/config pertaining to when and why should live elsewhere.

We utilise webhooks today for contract subscription, where the flow is to first create a webhook, and then setting up a subscription (this is where the business logic/config resides, like which contract to listen to, which events to index etc).

A similar pattern could be used here. We create a new `balance-subscriptions` table and endpoint. Logic/config related to the subscription should reside here (we can even expand later beyond this original scope, adding things like tracking erc20 balances). To make it general, we could have the `threshold` subscription be an optional configuration, default behaviour being we report any balance changes. IMO config should not be inside the webhook table.

nischitpra:
The main idea of having config in webhook table it to prevent creating multiple tables. Each webhook already has a type which can be used to perform specific tasks and should they require additional configurability, the config column helps specify it. It is an optional `jsonb` so basicaly the content can be anything so works for future webhooks as well. 

I had a discussing with @arcoraven before doing it this way as I wasn't sure if I should create a new table just for the config or just add it to the existing table.

